PR #37, opened 2022-04-27
packaging: fix rpm name

kmichel-aiven:

- fix support for / from tag name
- fix RPM name not prefixed with python3-


raphlopez (approved):
LGTM

PR #39, opened 2022-04-27
add more tests

carobme:
Add some more tests. Partially copied from pghoard.

This bumps coverage by 17%.


test/test_rohmufile.py
```diff
@@ -0,0 +1,54 @@
...
+from rohmu import rohmufile
+from rohmu.errors import InvalidConfigurationError
+from tempfile import NamedTemporaryFile
+
+
+def test_fileobj_name(tmpdir):
+    with NamedTemporaryFile(dir=tmpdir, suffix="foo") as raw_output_obj:
+        result = rohmufile._fileobj_name(raw_output_obj)  # pylint: disable=protected-access
+        assert result.startswith("open file ")
+        assert "foo" in result
```
alexole, line 15: The test for this method looks a bit artificial, as this is private and seems to be used only in two places and both are logging.

carobme: You're right that this test is definitely artificial.

I started from the top in `rohmufile.py` during dev-day and had no better idea what to test here. I don't think it matters that it's private (it's only by convention anyway) or that it's currently only called in two places. The main reason I created a test is that we don't always need to think about why some coverage is missing here (as our goal is 100%) and I thought adding a test - as stupid as it is - is still better than e.g. creating some document explaining why something isn't covered or adding something like `pragma: no cover` to the function.

I can see some value in making sure an instance of `NamedTemporaryFile` (or in general: an object with a `name` attribute) is accepted. Still, the asserts probably make no sense here. Maybe replace those with `assert isinstance(result, str)` (although that just dynamically checks what could be checked statically with mypy)? Or I could just drop the test completely.

alexole (approved):

PR #43, opened 2022-04-27
Fix automatic discovery in order to add subpackages distributed

axeoman:


alexole:
Could you resolve the conflicts please? @axeoman 
BTW instead of removing `include` we can just leave it as in the main now `include = rohmu*`, it does seem to work properly.

axeoman:
Yeah, I think this PR is leftover and we do not really need to merge it. Closing.

PR #51, opened 2022-05-17
misc: switch from unify/yapf to black

kmichel-aiven:

Replace unify/yapf formatter.

# Why this way

This makes all our repositories use the same tools.


pyproject.toml
```diff
@@ -1,3 +1,14 @@
...
+[tool.black]
+line-length = 125
+
+[tool.isort]
+no_sections = true
+force_alphabetical_sort = true
+combine_as_imports = true
+profile = "black"
+skip_gitignore = true
+line_length = 125
```
alexole, line 14: New line is missing

alexole (approved):
LGTM

PR #53, opened 2022-05-31
Delta: fail backup if required files are missing [BF-1203]

alexole:
Add support for required backup files, in case if file is missing the backup should fail.

# Why this way
This is the simplest approach, as file might disappear at any point in time. This solution is not very memory efficient as it needs to keep in memory all the required paths while doing snapshot, but it should suffice for the time being, for around 1 million of paths, it would take around ~10-100 megabytes, but I guess this snapshotting part in general might be improved in terms of memory efficiency.

### DIFF
--- a/rohmu/delta/snapshot.py
+++ b/rohmu/delta/snapshot.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2021 Aiven, Helsinki, Finland. https://aiven.io/
 from pathlib import Path
 from rohmu.delta.common import (
+    BackupPath,
     EMBEDDED_FILE_SIZE,
     hash_hexdigest_readable,
     increase_worth_reporting,
@@ -10,7 +11,7 @@ from rohmu.delta.common import (
     SnapshotHash,
     SnapshotState,
 )
-from typing import Callable, Dict, List, Optional
+from typing import Callable, Dict, Generator, Iterable, List, Optional, Sequence, Set, Tuple, Union
 
 import base64
 import logging
@@ -47,9 +48,9 @@ class Snapshotter:
         src,
         dst,
         globs,
-        src_iterate_func: Optional[Callable] = None,
-        parallel: Optional[int] = 1,
-        min_delta_file_size: int = 0
+        src_iterate_func: Optional[Callable[[], Iterable[Union[BackupPath, str, Path]]]] = None,
+        parallel: int = 1,
+        min_delta_file_size: int = 0,
     ):
         assert globs
         self.src = Path(src)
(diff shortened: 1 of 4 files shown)


rohmu/delta/snapshot.py
```diff
@@ -74,38 +75,58 @@ def _list_files(self, basepath: Path):
...
+            relative_path=relative_path,
+            mtime_ns=st.st_mtime_ns,
+            file_size=st.st_size,
+            stored_file_size=0,
+            missing_ok=missing_ok,
+        )
 
-    def _gen_snapshot_hashes(self, relative_paths, reuse_old_snapshotfiles):
+    def _gen_snapshot_hashes(
+        self, relative_paths: Sequence[Path], reuse_old_snapshotfiles: bool, required_paths: Optional[Set[Path]] = None
```
alanfranz-at-work, line 109: nit:
do we need required_paths to be an optional set? Can't we make this just a set, and if it's empty... it's empty? (we can use a frozenset as default) I think we could remove some ifs down there. (if required paths is an empty set, missing_files will always be an empty set, et cetera).

I think this will make the code simpler and we don't need to check/cast at every line.

rohmu/delta/snapshot.py
```diff
@@ -74,38 +75,58 @@ def _list_files(self, basepath: Path):
...
+    ) -> Generator[SnapshotFile, None, None]:
         same = 0
         lost = 0
+
+        if required_paths:
```
alanfranz-at-work, line 114: Are the paths guaranteed to have the same semantics, or required_paths could be absolute while relative_paths could be... relative? AFAIK in pathlib that if p1.absolute() == p2.absolute(), then p1 doesn't necessarily == p2 if they have a different representation. If they could have a different representation we'd need to normalize them somehow before doing those checks.

ghost (changes requested):
Mostly LGTM, one question to address (without necessarily changing code), and minor nitpicking (that you may choose to ignore).

rohmu/delta/snapshot.py
```diff
@@ -47,9 +48,9 @@ def __init__(
         src,
         dst,
         globs,
-        src_iterate_func: Optional[Callable] = None,
-        parallel: Optional[int] = 1,
-        min_delta_file_size: int = 0
+        src_iterate_func: Optional[Callable[[], Iterable[Union[BackupPath, str, Path]]]] = None,
```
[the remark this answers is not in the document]
alexole: Yes

rohmu/delta/snapshot.py
alexole: I agree to an extent, I don't think it will make code much simpler, e.g. this check will not go away
```
        if required_paths:
            missing_files = required_paths.difference(set(relative_paths))
```
as I don't want to waste resources on transforming paths to set and calculating the difference between sets if nothing is required.

ghost (approved):
LGTM

rohmu/delta/snapshot.py
alexole: Well, it's a small thing, but what I meant is that `required_paths.difference(set(relative_paths))` in this case will need to be done even if there is no required files, which actually means transforming relative_paths to set for no reason, if the list is big - unnecessary extra memory usage.

PR #58, opened 2022-07-21
allow passing timeouts to s3 wrapper

sjamgade:
forwards timeouts to the botocore lib for better control, otherwise the process just hangs

Resolves: #xxxxx

# Why this way

### DIFF
--- a/rohmu/object_storage/s3.py
+++ b/rohmu/object_storage/s3.py
@@ -62,14 +62,21 @@ class S3Transfer(BaseTransfer):
         segment_size=MULTIPART_CHUNK_SIZE,
         encrypted=False,
         proxy_info=None,
+        connect_timeout=None,
+        read_timeout=None,
     ):
         super().__init__(prefix=prefix)
         botocore_session = botocore.session.get_session()
         self.bucket_name = bucket_name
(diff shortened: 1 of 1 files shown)


rohmu/object_storage/s3.py
```diff
@@ -62,14 +62,21 @@ def __init__(
...
         super().__init__(prefix=prefix)
         botocore_session = botocore.session.get_session()
         self.bucket_name = bucket_name
         self.location = ""
         self.region = region
+        timeouts = {}
+        if connect_timeout:
+            timeouts["connect_timeout"] = connect_timeout
+        if read_timeout:
+            timeouts["read_timeout"] = read_timeout
```
hackaugusto, line 77: > read_timeout (float or int) -- The time in seconds till a timeout exception is thrown when attempting to read from a connection. The default is 60 seconds.

> connect_timeout (float or int) -- The time in seconds till a timeout exception is thrown when attempting to make a connection. The default is 60 seconds.

https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html#botocore.config.Config

hackaugusto (approved):
LGTM

PR #59, opened 2022-08-22
 teach rohmu to notifiy when it writes anything [AD-756]

sjamgade:
This allows robjecter to be informed everytime pghoard (rohmu) writes
anything. This can be controlled with the presence of extra config
passed to the constructors of each object storage driver.

The plain_key is needed as robjecter itself uses rohmu so the prefix
gets prefixed twice.
size is not strictly needed but helps make robjecter's work easier. And
here size is easily available statically or dynamically (in case of
multipart-uploads). In case of multipart uploads the size could
calculated using the callback mechanisms locally however each driver
reports callback in a different manner and in some cases the callback
might not be called if a multipart upload was not necessary.

NOTE: this just replays aiven/pghoard/pull/548 here 

[AD-756]

Resolves: #xxxxx

# Why this way

[AD-756]: https://aiven.atlassian.net/browse/AD-756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rohmu/object_storage/azure.py
+++ b/rohmu/object_storage/azure.py
@@ -11,6 +11,7 @@ from io import BytesIO
 
 import azure.common
 import logging
+import os
 import time
 
 try:
@@ -191,6 +192,7 @@ class AzureTransfer(BaseTransfer):
                 )
 
     def delete_key(self, key):
+        plain_key = key
         key = self.format_key_for_backend(key, remove_slash_prefix=True)
         self.log.debug("Deleting key: %r", key)
         try:
@@ -198,6 +200,7 @@ class AzureTransfer(BaseTransfer):
             return blob_client.delete_blob()
         except azure.core.exceptions.ResourceNotFoundError as ex:  # pylint: disable=no-member
             raise FileNotFoundFromStorageError(key) from ex
+        self.notify_delete(plain_key)
 
     def get_contents_to_file(self, key, filepath_to_store_to, *, progress_callback=None):
         key = self.format_key_for_backend(key, remove_slash_prefix=True)
@@ -288,6 +291,7 @@ class AzureTransfer(BaseTransfer):
     def store_file_from_memory(self, key, memstring, metadata=None, cache_control=None, mimetype=None):
         if cache_control is not None:
             raise NotImplementedError("AzureTransfer: cache_control support not implemented")
+        plain_key = key
         key = self.format_key_for_backend(key, remove_slash_prefix=True)
         content_settings = None
         if mimetype:
@@ -300,10 +304,12 @@ class AzureTransfer(BaseTransfer):
             metadata=self.sanitize_metadata(metadata, replace_hyphen_with="_"),
             overwrite=True,
         )
+        self.notify_write(key=plain_key, size=len(bytes(memstring)))
 
     def store_file_from_disk(self, key, filepath, metadata=None, multipart=None, cache_control=None, mimetype=None):
         if cache_control is not None:
             raise NotImplementedError("AzureTransfer: cache_control support not implemented")
+        plain_key = key
         key = self.format_key_for_backend(key, remove_slash_prefix=True)
         content_settings = None
         if mimetype:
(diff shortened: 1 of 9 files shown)


sjamgade:
(Assigned reviewers based on github's recommendations :-) )

rohmu/object_storage/s3.py
```diff
@@ -395,6 +402,7 @@ def multipart_upload_file_object(self, *, cache_control, fp, key, metadata, mime
         self.log.info(
             "Multipart upload of %r complete, size: %r, took: %.2fs", key, size, time.monotonic() - start_of_multipart_upload
         )
+        self.notify_write(key=key, size=self.get_file_size(plain_key))
```
ngilles-aiven, line 405: `get_file_size()` could fail here (also true for other remote storage drivers), should we handle that and allow a value for unknown file size (`-1` ?)

rohmu/object_storage/azure.py
```diff
@@ -191,13 +192,15 @@ def _iter_key(self, *, path, with_metadata, deep):
...
     def delete_key(self, key):
+        plain_key = key
         key = self.format_key_for_backend(key, remove_slash_prefix=True)
         self.log.debug("Deleting key: %r", key)
         try:
             blob_client = self.conn.get_blob_client(container=self.container_name, blob=key)
             return blob_client.delete_blob()
         except azure.core.exceptions.ResourceNotFoundError as ex:  # pylint: disable=no-member
             raise FileNotFoundFromStorageError(key) from ex
+        self.notify_delete(plain_key)
```
ngilles-aiven, line 203: feels odd that these common behaviours of notification needs to be explicitly added to every provider/implementation... could this be extracted?

rohmu/object_storage/base.py
```diff
@@ -121,6 +139,36 @@ def store_file_from_disk(self, key, filepath, metadata=None, multipart=None, cac
...
+        if self.notification_url:
+            self.notifier.submit(self._notify, operation)
+
+    def _notify(self, operation):
+        """Actually notify, this gets eventually called from the pool"""
+        try:
+            with get_requests_session() as session:
+                session.post(f"{self.notification_url}", json=operation)
+        except requests.RequestException as ex:
+            self.log.warning("Could not inform robjecter about %r, because %s", operation, ex)
```
ngilles-aiven, line 157: Does `rohmu` really need to know or mention robjecter explicitly? Shouldn't we consider this a generic webhook implementation?

rohmu/object_storage/azure.py
hackaugusto: how would you go about extracting it?

hackaugusto:
Documenting the Slack discussion. The decision was to go with composition.

## Inheritance

Extend the BaseTransfer class to include notification

Pros:
- Has all the details about the upload available

Cons:
- Code increases with the number of base implementations
  - Duplicated code, i.e. every subclass must reimplement the same logic
  - More error prone, i.e. forget to add it in a base class
  - More tests, i.e. because every subclass reimplements the same logic, they have to be tested separately
- Mixes the concerns of notification with transfer

## Proxy class

Write a proxy class to call notify.

Pros:

- Single class implementing notify
  - no duplicated logic 
  - fewer tests

Cons:
- A näive implementation doesn't have all the data available, so additional calls are required. This includes OS system calls (cheap), and REST calls like `store_file_object` used for streaming uploads

## Composition

Pass a notifier object to the base transfer

Pros:
- Isolate the responsabilities
- Allows the reuse of the data for streaming uploads

Cons:
- Same cons as inheritance w.r.t. number of implementations

rohmu/object_storage/azure.py
hackaugusto: FYI: This was considered an non-issue.

rohmu/object_storage/s3.py
hackaugusto: Do you mean if the cloud provider has consistency issues? (Well, that is probably the case for Azure, their API is weirdly inconsistent)

ngilles-aiven: Also, but even just the fact it's ann HTTP HEAD request... chances of it failing should be quite low (especially if the transfer goes through :D) but still, it "could" (transient network hiccup)

hackaugusto:
superseeded by https://github.com/aiven/rohmu/pull/67

PR #60, opened 2022-09-06
Make sure get_contents_to_file() calls are atomic

ghost:
Make sure that calls to get_contents_to_file() are actually atomic

Calls to get_contents_to_file() weren't necessarily atomic - destination file
could be visible before it was completely written, or it could be left half-written
on the filesystem if the process crashed, or it could leave other temporary files
that nothing would cleanup in the same directory as target file.

This was related to some bugs that can be seen in pghoard.

We fix all these issues through an anonymous temporary file approach; details
in commit messages.

### DIFF
--- a/rohmu/atomic_opener.py
+++ b/rohmu/atomic_opener.py
@@ -1,18 +1,31 @@
 from contextlib import contextmanager
 from pathlib import Path
-from typing import (
-    BinaryIO,
-    Callable,
-    cast,
-    ContextManager,
-    Generator,
-    Literal,
-    Optional,
-    overload,
-    TextIO,
-    TYPE_CHECKING,
-    Union,
-)
+from typing import BinaryIO, Callable, cast, ContextManager, Generator, Optional, overload, TextIO, TYPE_CHECKING, Union
+
+# workaround < Python3.8 missing Literal.
+# if we just use the try/except block, mypy will complain
+# about the fact that Write* variables are being redefined.
+# so we need to pour in an additional branch.
+if TYPE_CHECKING:
+    from typing import Literal
+
+    Write = Literal["w"]
+    WriteBinary = Literal["wb"]
+    WriteTest = Literal["somethingrandomw"]
+else:
+    try:
+        from typing import Literal
+
+        Write = Literal["w"]
+        WriteBinary = Literal["wb"]
+        WriteTest = Literal["somethingrandomw"]
+    except ImportError:
+        from typing import Any
+
+        Write = Any
+        WriteBinary = Any
+        WriteTest = Any
+
 
 import errno
 import os
@@ -31,7 +44,7 @@ def _fd_close_quietly(fd: int) -> None:
 @overload
 def atomic_opener(
     final_path: Path,
-    mode: Literal["wb"],
+    mode: WriteBinary,
     encoding: Optional[str] = None,
     _fd_spy: Callable[[int], None] = lambda unused: None,
     _after_link_hook: Callable[[], None] = lambda: None,
@@ -42,7 +55,7 @@ def atomic_opener(
 @overload
 def atomic_opener(
     final_path: Path,
-    mode: Literal["w"],
+    mode: Write,
     encoding: Optional[str] = None,
     _fd_spy: Callable[[int], None] = lambda unused: None,
     _after_link_hook: Callable[[], None] = lambda: None,
@@ -55,7 +68,7 @@ if TYPE_CHECKING:
     @overload
     def atomic_opener(
         final_path: Path,
-        mode: Literal["somethingrandomw"],
+        mode: WriteTest,
         encoding: Optional[str] = None,
         _fd_spy: Callable[[int], None] = lambda unused: None,
         _after_link_hook: Callable[[], None] = lambda: None,
--- a/rohmu/atomic_opener.py
+++ b/rohmu/atomic_opener.py
@@ -1,0 +1,115 @@
+from contextlib import contextmanager
+from pathlib import Path
+from typing import (
+    BinaryIO,
+    Callable,
+    cast,
+    ContextManager,
+    Generator,
+    Literal,
+    Optional,
+    overload,
+    TextIO,
+    TYPE_CHECKING,
+    Union,
+)
+
+import errno
+import os
+
+
+def _fd_close_quietly(fd: int) -> None:
+    try:
+        os.close(fd)
+    except OSError as e:
+        if e.errno == errno.EBADF:
+            # closed already
+            return
+        raise
+
+
+@overload
+def atomic_opener(
+    final_path: Path,
+    mode: Literal["wb"],
+    encoding: Optional[str] = None,
+    _fd_spy: Callable[[int], None] = lambda unused: None,
+    _after_link_hook: Callable[[], None] = lambda: None,
+) -> ContextManager[BinaryIO]:
+    ...
+
+
+@overload
+def atomic_opener(
+    final_path: Path,
+    mode: Literal["w"],
+    encoding: Optional[str] = None,
+    _fd_spy: Callable[[int], None] = lambda unused: None,
+    _after_link_hook: Callable[[], None] = lambda: None,
+) -> ContextManager[TextIO]:
+    ...
+
+
+if TYPE_CHECKING:
+    # necessary for testing or tests will fail type checking
+    @overload
+    def atomic_opener(
+        final_path: Path,
+        mode: Literal["somethingrandomw"],
+        encoding: Optional[str] = None,
+        _fd_spy: Callable[[int], None] = lambda unused: None,
+        _after_link_hook: Callable[[], None] = lambda: None,
+    ) -> ContextManager[TextIO]:
+        ...
+
+
+def atomic_opener(
+    final_path: Path,
+    mode: str,
+    encoding: Optional[str] = None,
+    _fd_spy: Callable[[int], None] = lambda unused: None,
+    _after_link_hook: Callable[[], None] = lambda: None,
+) -> ContextManager[Union[TextIO, BinaryIO]]:
+    return _atomic_opener(final_path, mode, encoding, _fd_spy, _after_link_hook)
+
+
+@contextmanager
+def _atomic_opener(
+    final_path: Path,
+    mode: str,
+    encoding: Optional[str] = None,
+    _fd_spy: Callable[[int], None] = lambda unused: None,
+    _after_link_hook: Callable[[], None] = lambda: None,
+) -> Generator[Union[TextIO, BinaryIO], None, None]:
+    """
+    Creates a file object for writing which will only appear on the filesystem if the context manager succeeds.
+
+    It's designed to prevent partial writes (file will either be available with full content, or will be unavailable)
+    and the need for manual cleanup.
+    """
+    parent_dir = final_path.parent
+    if not parent_dir.exists():
+        raise IOError(f"Parent directory '{parent_dir}' must exist but is missing")
+    if "w" not in mode:
+        raise ValueError("Write mode must be used to make actual sense")
+
+    fd = os.open(str(parent_dir), os.O_TMPFILE | os.O_RDWR, 0o600)
+    _fd_spy(fd)
+    try:
+        file_obj = os.fdopen(fd, mode=mode, encoding=encoding)
+    except Exception:  # pylint: disable=broad-except
+        # when passing wrong mode, os.fdopen won't close input fd. When passing wrong encoding, it will.
+        _fd_close_quietly(fd)
+        raise
+
+    try:
+        if "b" in mode:
+            yield cast(BinaryIO, file_obj)
+        else:
+            yield cast(TextIO, file_obj)
+        file_obj.flush()
+        path = "/proc/self/fd/{0}".format(fd)
+        os.link(path, str(final_path), src_dir_fd=0, follow_symlinks=True)
+        _after_link_hook()
+    finally:
+        file_obj.close()
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,7 +57,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.9, 3.8, 3.7, 3.6]
+        python-version: ["3.10", "3.9", "3.8", "3.7"]
 
     steps:
       - id: checkout-code
--- a/rohmu/object_storage/azure.py
+++ b/rohmu/object_storage/azure.py
@@ -8,6 +8,8 @@ See LICENSE for details
 from azure.core.exceptions import HttpResponseError, ResourceExistsError
 from azure.storage.blob import BlobServiceClient, ContentSettings
 from io import BytesIO
+from pathlib import Path
+from rohmu.atomic_opener import atomic_opener
 
 import azure.common
 import logging
@@ -204,7 +206,7 @@ class AzureTransfer(BaseTransfer):
 
         self.log.debug("Starting to fetch the contents of: %r to: %r", key, filepath_to_store_to)
         try:
-            with open(filepath_to_store_to, "wb") as f:
+            with atomic_opener(Path(filepath_to_store_to), mode="wb") as f:
                 container_client = self.conn.get_container_client(self.container_name)
                 container_client.download_blob(key).readinto(f)
         except azure.core.exceptions.ResourceNotFoundError as ex:  # pylint: disable=no-member
--- a/rohmu/object_storage/google.py
+++ b/rohmu/object_storage/google.py
@@ -5,7 +5,6 @@ Copyright (c) 2016 Ohmu Ltd
 See LICENSE for details
 """
 # pylint: disable=import-error, no-name-in-module
-
 from ..dates import parse_timestamp
 from ..errors import FileNotFoundFromStorageError, InvalidConfigurationError
 from .base import BaseTransfer, get_total_memory, IterKeyItem, KEY_TYPE_OBJECT, KEY_TYPE_PREFIX
@@ -14,9 +13,11 @@ from googleapiclient.discovery import build
 from googleapiclient.errors import HttpError
 from googleapiclient.http import build_http, MediaFileUpload, MediaIoBaseDownload, MediaIoBaseUpload, MediaUpload
 from http.client import IncompleteRead
-from io import BytesIO, FileIO
+from io import BytesIO
 from oauth2client import GOOGLE_TOKEN_URI
 from oauth2client.client import GoogleCredentials
+from pathlib import Path
+from rohmu.atomic_opener import atomic_opener
 
 import codecs
 import errno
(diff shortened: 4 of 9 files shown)


rohmu/atomic_open.py
```diff
@@ -0,0 +1,40 @@
...
+        raise IOError(f"Parent directory '{parent_dir}' must exist but is missing")
+    if "w" not in mode:
+        raise ValueError("Write mode must be used to make actual sense")
+
+    fd = os.open(str(parent_dir), os.O_TMPFILE | os.O_RDWR, 0o600)
+    file_obj = os.fdopen(fd, mode=mode, encoding=encoding)
+    try:
+        yield file_obj
+        path = "/proc/self/fd/{0}".format(fd)
+        _inode_spy(os.stat(path).st_ino)
```
kmichel-aiven, line 37: You can get the `fd` from the `file_obj` by doing  `file_obj.fileno()` and then you can call `os.stat` of that fd directly without using the path:
 - No need to construct and know about the linux `/proc/self/...` path 
 - No need for a test-specific hook since you already can reach the `fd` from the `file_obj`

test/test_atomic_open.py
```diff
@@ -0,0 +1,107 @@
...
+            finished_writing_file_at.append(time.monotonic())
+            should_stop.set()
+            # we make sure not to trigger a race condition with check_not_created
+            time.sleep(wait_after_write_seconds)
+
+        if not finished_writing_file_at:
+            raise ValueError("didn't finish writing")
+        # this must be super-fast since atomic move is taking place. Makes more sense with
+        # a large - 100ish - block count
+        assert (time.monotonic() - finished_writing_file_at[0] - wait_after_write_seconds) <= 0.1
```
kmichel-aiven, line 50: This feels *very* flaky, since it does not really prove atomicity, I'm not sure it's worth it.

test/test_atomic_open.py
```diff
@@ -0,0 +1,107 @@
...
+        assert not output_file.exists()
+        tmpfile_inode = []
+        with atomic_opener(output_file, _inode_spy=tmpfile_inode.append, mode="w") as f:
+            for unused_counter in range(block_count):
+                f.write(data_block)
+            time.sleep(3)
+            finished_writing_file_at.append(time.monotonic())
+            should_stop.set()
+            # we make sure not to trigger a race condition with check_not_created
+            time.sleep(wait_after_write_seconds)
```
kmichel-aiven, line 44: This feels flaky, looks like waiting for `future.result()` would achieve the same result.

test/test_atomic_open.py
```diff
@@ -0,0 +1,107 @@
...
+    finished_writing_file_at = []
+    wait_after_write_seconds = 0.5
+
+    try:
+        assert not output_file.exists()
+        tmpfile_inode = []
+        with atomic_opener(output_file, _inode_spy=tmpfile_inode.append, mode="w") as f:
+            for unused_counter in range(block_count):
+                f.write(data_block)
+            time.sleep(3)
```
kmichel-aiven, line 40: Maybe better to wait until a certain number of loops of `check_not_created` have run, with a counter, there is no real proof that the `check_not_created` thread made any progress during that sleep.

kmichel-aiven (dismissed):

test/test_atomic_open.py
alanfranz-at-work: It doesn't prove atomicity with a small file, but I didn't feel putting a very large write in an automated tests, since it could be unnecessarily slow.

rohmu/atomic_open.py
```diff
@@ -0,0 +1,40 @@
...
+    It's designed to prevent partial writes (file will either be available with full content, or will be unavailable)
+    and the need for manual cleanup.
+    """
+    parent_dir = final_path.parent
+    if not parent_dir.exists():
+        raise IOError(f"Parent directory '{parent_dir}' must exist but is missing")
+    if "w" not in mode:
+        raise ValueError("Write mode must be used to make actual sense")
+
+    fd = os.open(str(parent_dir), os.O_TMPFILE | os.O_RDWR, 0o600)
```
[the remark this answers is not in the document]
alanfranz-at-work: well, even if it did (but it throws an exception in Python nevertheless AFAICT) when using O_TMPFILE the semantics say that the file is deleted as soon as no fd references it. If it fails, no fd will reference this, and the file will be deleted.

rohmu/atomic_open.py
```diff
@@ -0,0 +1,40 @@
...
+
+import os
+
+
+def atomic_open_for_writing_text(final_path: Path, encoding: Optional[str] = None) -> ContextManager[TextIO]:
+    return cast(ContextManager[TextIO], atomic_opener(final_path, encoding=encoding, mode="w"))
+
+
+def atomic_open_for_writing_bytes(final_path: Path) -> ContextManager[BinaryIO]:
+    return cast(ContextManager[BinaryIO], atomic_opener(final_path, mode="wb"))
```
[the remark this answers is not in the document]
alanfranz-at-work: Why do you think they're dangerous in this context? I'm using an hardcoded mode literal.

I'm checking the literal-dependent typing, but that feels much more mage to me than what I did (python open() returning object that can handle either bytes or str is one of the bad design choices in Python imho).

rohmu/atomic_open.py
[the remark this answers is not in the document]
kmichel-aiven: It throws an exception, but outside of the finally.
I doubt "as soon as no fd will reference this" applies to the `fd` var, since it's an int, so Python won't close the file descriptor when fd-the-var leaves the scope).
And I don't think this applies to "the abstract thing pointed to by fd", since it would allow linux/python to close the file descriptor even before `fdopen` is executed over the file descriptor. That abstract thing has no variable holding onto it.
So that would mean this applies to the end of the program, which means progressive leaking of fd in long running programs.

test/test_atomic_open.py
kmichel-aiven: I think it's the best you can do :)

But the false-negative combined with the false-positive of "python is sometimes slower than expected during the tests, so sometimes we have more than 0.1s between two lines of code" (becaues of swap, slow print io, other thread keeping the GIL, coverage instrumentation....), makes me think maybe it's not worth attempting to test that at all.

rohmu/atomic_open.py
alanfranz-at-work: This one I'm not understanding. I need the ```/proc/self/...``` to create the hardlink to the actual filesystem, this is straight from manpage:

```
 snprintf(path, PATH_MAX,  "/proc/self/fd/%d", fd);
                  linkat(AT_FDCWD, path, AT_FDCWD, "/path/for/file",
                                          AT_SYMLINK_FOLLOW);
```

Also, I need the test-specific hook because I want to make sure the inode from the created file is actually the same of the file which gets created inside the block. This is to verify that a copy doesn't happen, since it's the same inode.

I'm not sure how you'd use the fd number to perform the same check. File descriptors are process-local and there's no guarantee of non-reuse AFAIK and, if they were, I'd risk testing implementation details rather than the interface.

test/test_atomic_open.py
alanfranz-at-work: That's right, good idea, thanks!

rohmu/atomic_open.py
kmichel-aiven: I forgot about the `path` in the link :clown_face: So yes, the path has to stay.

However, for the test, what I meant is in your test you can do:
```python
    with atomic_open_for_writing_bytes(final_path) as tmp_file:
        tmp_inode = os.stat(tmp_file.fileno()).st_ino
   assert tmp_inode == os.stat(final_path).st_ino
```
You're still comparing the inode, not comparing the fd.

rohmu/atomic_open.py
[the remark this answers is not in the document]
kmichel-aiven: They are dangerous because the cast function does verify the type of it's input at all,
so if `atomic_opener` is later updated to return a different type, the cast will hide the error.

It's only correct right now, assuming the rest of the code doesn't change,
which isn't in the spirit of relying on type to avoid errors.

(I agree about Python and the mixed bytes/str output, heritage of the pre-typing-times, and overloads are annoying and verbose to type, but users of libraries > writers of libraries).

rohmu/atomic_open.py
```diff
@@ -0,0 +1,40 @@
...
+
+    fd = os.open(str(parent_dir), os.O_TMPFILE | os.O_RDWR, 0o600)
+    file_obj = os.fdopen(fd, mode=mode, encoding=encoding)
+    try:
+        yield file_obj
+        path = "/proc/self/fd/{0}".format(fd)
+        _inode_spy(os.stat(path).st_ino)
+        os.link(path, str(final_path), src_dir_fd=0, follow_symlinks=True)
+    finally:
+        file_obj.close()
```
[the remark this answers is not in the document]
alanfranz-at-work: MMMh. Let's see if I can test this. Good catch.

rohmu/atomic_open.py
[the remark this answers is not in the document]
alanfranz-at-work: Mmmh, maybe github is highlighting the wrong lines so my previous answer is unrelated.

Your concern is about os.fdopen failing when os.open has succeded, so we actually have an open file descriptor already.

I'd say that os.fdopen shouldn't throw, but we can't know, so yeah, we could add a check.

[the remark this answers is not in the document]
alanfranz-at-work: Do you know a fun things?

os.fdopen may (if wrong encoding) or may not (if wrong mode) close the file descriptor which is passed in :-/ so we need to check what happens.

rohmu/atomic_open.py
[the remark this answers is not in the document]
alanfranz-at-work: I don't seem to get this to actually work. I get this error:

```
rohmu/atomic_open.py:45: error: Overloaded function implementation cannot produce return type of signature 1
rohmu/atomic_open.py:45: error: Overloaded function implementation cannot produce return type of signature 2
```
Pycharm doesn't seem to complain, though, and the type chain seems correct to me.

If I need to cast file_obj from IO[Any] to the proper IO subtype, we're back to casting and I don't know whether this is useful.

rohmu/atomic_open.py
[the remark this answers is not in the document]
alanfranz-at-work: It was a potential bug after all, solved. Thanks!

rohmu/atomic_opener.py
```diff
@@ -0,0 +1,128 @@
...
+    and the need for manual cleanup.
+    """
+    parent_dir = final_path.parent
+    if not parent_dir.exists():
+        raise IOError(f"Parent directory '{parent_dir}' must exist but is missing")
+    if "w" not in mode:
+        raise ValueError("Write mode must be used to make actual sense")
+
+    fd = os.open(str(parent_dir), os.O_TMPFILE | os.O_RDWR, 0o600)
+    _fd_spy(fd)
```
kmichel-aiven, line 110: From previous conversation, you don't need _fd_spy at all, you can read file_obj.fileno() of the temp file during the tests.

kmichel-aiven (dismissed):

rohmu/atomic_opener.py
alanfranz: @kmichel-aiven I apologize, I should have added a note because this was difficult to notice. 

I had removed and then re-added it because, in fact, now I do: there're the file descriptor leak tests that actually check whether the fd is still open even when the open failed; I need the fd in those, since the python-level file object is never opened successfully.

rohmu/atomic_opener.py
```diff
@@ -0,0 +1,128 @@
...
+# about the fact that Write* variables are being redefined.
+# so we need to pour in an additional branch.
+if TYPE_CHECKING:
+    from typing import Literal
+
+    Write = Literal["w"]
+    WriteBinary = Literal["wb"]
+    WriteTest = Literal["somethingrandomw"]
+else:
+    try:
```
hackaugusto, line 16: This `try` is probably unecessary. The types can be set to `Any` if it is not type checking

rohmu/atomic_opener.py
```diff
@@ -0,0 +1,128 @@
...
+    It's designed to prevent partial writes (file will either be available with full content, or will be unavailable)
+    and the need for manual cleanup.
+    """
+    parent_dir = final_path.parent
+    if not parent_dir.exists():
+        raise IOError(f"Parent directory '{parent_dir}' must exist but is missing")
+    if "w" not in mode:
+        raise ValueError("Write mode must be used to make actual sense")
+
+    fd = os.open(str(parent_dir), os.O_TMPFILE | os.O_RDWR, 0o600)
```
hackaugusto, line 109: why not use the standard library like it was done with `atomic_create_file`?

It also fixes issues with weird `O_TMPFILE`:

> The above constants are extensions and not present if they are not defined by the C library.

Which would be the case for:

> [..] platforms that are neither Posix nor Cygwin

alanfranz: atomic_create_file() is indeed easier, but it pollutes source directory with .tmp files with no chance of automated cleanup, so it requires the caller to be 100% aware of what's happening and perform compensative actions if complete failures (e.g process crash) happen.

> The above constants are extensions and not present if they are not defined by the C library.

It's available since 9 years in the Linux kernel. Does pghoard support anything else than Linux, actually?

rohmu/atomic_opener.py
alanfranz: Probably. And this will fail if using mypy on Python <= 3.8, which we don't do on our CI.

I personally dislike the Literal typing approach (too complex) but I was asked to do this. 

Is removing the try/except any better, though? I think the else block is a clear, complete path for when we don't use TYPE_CHECKING is unavailable, while TYPE_CHECKING is the exceptional Path that I had to add because of some mypy warts.

alanfranz: Possibly, but I really think too much time was wasted on typing here. The original code used a cast, that's true, but it was internal to a module. I think that type checks should help us, not make our life more difficult.

rohmu/atomic_opener.py
hackaugusto: Thank you for the explanation. Indeed `mkstemp` doesn't delete the file:

> Unlike [TemporaryFile()](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile), the user of [mkstemp()](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp) is responsible for deleting the temporary file when done with it.

I also tested with:

```python
import tempfile, sys, os.path

file_path = "/tmp/test"
fd, tmp_file_path = tempfile.mkstemp(
    prefix=os.path.basename(file_path),
    dir=os.path.dirname(file_path),
    suffix=".metadata_tmp",
)
print(tmp_file_path)
sys.exit(0)  # above file is not deleted!
```

But maybe we can still use `NamedTemporaryFile` from the stdlib? (No strong feelings, now the work has been done anyways, it would just have saved a bit of work with the typing stuff)

```python
import tempfile, sys, os

with tempfile.NamedTemporaryFile() as f:
    print(f.name)
    os.link(f.name, "/tmp/test")
    sys.exit(0)  # temp file is gone, /tmp/test is created
```

> It's available since 9 years in the Linux kernel. Does pghoard support anything else than Linux, actually?

Well, I was just pointing out the advantage listed on the Python docs, and seem easy enough to use. But yeah, I suppose it is not an explicit goal of any of our projects

rohmu/atomic_opener.py
hackaugusto: > Is removing the try/except any better, though?

Well, it is just a bit less code, but not significatly better. I guess your rationale is "lets try to make the types as close to the correct values as possible", whereas I would go with "lets do the minimum that works for type checking and runtime". No need to change it though :)

rohmu/atomic_opener.py
alanfranz: > But maybe we can still use `NamedTemporaryFile` from the stdlib?

sys.exit() is the graceful way to exit Python. It's just like assuming a "finally" block will always be executed. It's got exactly the same issues as mkstemp. What happens if the process exits abruptly? We're left with a file that nobody will ever cleanup - and, during the write operations, which could take time, a temporary file appears in the same directory as the destination - which is something that the client might not expect.

```
>>> with tempfile.NamedTemporaryFile() as f:
...    print(f.name)
...    os._exit(1)
... 
/tmp/tmpu_5wsslx
alan.franzoni@toolbox:~/aiven/aiven-core(dmitry-potepalov-cassandra-fix-rack-distribution)$ ls -l /tmp/tmpu_5wsslx
-rw------- 1 alan.franzoni alan.franzoni 0 Sep 14 16:21 /tmp/tmpu_5wsslx
```

I didn't do this because I wanted; I tried to find other ways.

In order to do this "the right way", we either use O_TMPFILE, or we need to remove ```get_contents_to_file``` from rohmu interface; then a client is forced to properly handle file behaviour as it sees fit. If you people think we should remove the method from the interface and the current implementation has issues, I'll go that way; but just trying to look where rohmu is actually used could take a bit of time, since it's public we should discuss deprecations, etc.

hackaugusto: > sys.exit() is the graceful way to exit Python.

I always forget it throws an exception 😬

alanfranz: @hackaugusto I tried with TemporaryFile:
```
    with TemporaryFile(dir=parent_dir, mode=mode, encoding=encoding) as file_obj:
        if "b" in mode:
            yield cast(BinaryIO, file_obj)
        else:
            yield cast(TextIO, file_obj)
        file_obj.flush()
        path = "/proc/self/fd/{0}".format(file_obj.fileno())
        assert Path(path).exists()
        os.link(path, str(final_path), src_dir_fd=0, follow_symlinks=True)
        _after_link_hook()
```

But this seems to fail:
```
            path = "/proc/self/fd/{0}".format(file_obj.fileno())
            assert Path(path).exists()
>           os.link(path, str(final_path), src_dir_fd=0, follow_symlinks=True)
E           FileNotFoundError: [Errno 2] No such file or directory: '/proc/self/fd/11' -> '/tmp/pytest-of-alan.franzoni/pytest-17/test_file_is_atomically_create0/something'
```

So, it seems we get a different set of problems by using python's own TemporaryFile, but I think this "atomic open" function would be a nice addition to python's own stdlib and I'll try contributing it upstream, if it's possible to make this work.

alanfranz: So, I tried investigating since I was at it; it seems that Python uses O_EXCL when opening a TemporaryFile, and this prevents (implementation choice) to actually link in the filesystem. I think this is more incidental than intentional, though.

hackaugusto: @alanfranz thanks for looking into it :bow:

kmichel-aiven:
This needs to be rebased.

alanfranz:
Moved over to https://github.com/aiven/rohmu/pull/69

PR #61, opened 2022-09-06
typing for init

hackaugusto:
Super small PR that introduces additional typing for the config object

### DIFF
--- a/rohmu/__init__.py
+++ b/rohmu/__init__.py
@@ -9,10 +9,11 @@ from .object_storage.base import BaseTransfer
 from typing import Type
 
 IO_BLOCK_SIZE = 2**20  # 1 MiB
+STORAGE_TYPE = "storage_type"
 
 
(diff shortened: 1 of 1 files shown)


rohmu/__init__.py
```diff
@@ -6,13 +6,15 @@
 """
 from .errors import InvalidConfigurationError
 from .object_storage.base import BaseTransfer
-from typing import Type
+from typing import Any, Dict, Type
 
 IO_BLOCK_SIZE = 2**20  # 1 MiB
+STORAGE_TYPE = "storage_type"
+Config = Dict[str, Any]
```
kmichel-aiven, line 13: This should take a `Mapping` instead of a `Dict`, to make it clear that the input parameter will not be modified.

kmichel-aiven (changes requested):

ngilles-aiven (approved):
LGTM

PR #62, opened 2022-09-06
No longer shadow the argument key

hackaugusto:
This is a biggish change, I have it in a separate PR so that reviewing can be easier.

It just makes sure the value `key` is no longer shadowed, so it consistently name the formatted key `path`. The original value of `key` will be used by the notifier interface (an alternative implementation of #59)


sjamgade (approved):
LGTM, will wait before merging if someone else wants to have a look

PR #64, opened 2022-09-09
drop support for Python 3.6 [EOL 23/Dec/2021]

hackaugusto:
As the tile says, 3.6 has reached EOL last year


ngilles-aiven (approved):
LGTM

PR #65, opened 2022-09-09
build: fix python version on yaml

hackaugusto:
The value `3.10` is parsed as an integer and the right most zero is dropped. The effect is that GH CI tried to run the build using python 3.1 instead of 3.10. The fix is to use strings to define the version instead of numbers.


sjamgade (approved):

hackaugusto:
@sjamgade can we merge this?

PR #66, opened 2022-09-09
Notifier implementation

hackaugusto:
Alternative implementation of https://github.com/aiven/rohmu/pull/59 using composition instead of inheritance.

This only adds the notifier interface and a few implementations, it does not yet integreate the code in the drivers (will be done on a follow up, split to make reviewing easier)

### DIFF
--- a/rohmu/notifier/http.py
+++ b/rohmu/notifier/http.py
@@ -1,0 +1,172 @@
+"""Copyright (c) 2022 Aiven, Helsinki, Finland. https://aiven.io/"""
+from .interface import Notifier
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from enum import Enum, unique
+from logging import getLogger
+from queue import Empty, Queue
+from requests.exceptions import RequestException
+from typing import Optional
+
+import json
+import requests
+import threading
+
+LOG = getLogger(__name__)
+
+
+# Notifiers can be a short-lived objects, the threads must quit so to not leak resources. This
+# timeout controls how often a thread will check for termination.
+_CHECK_STOP_EVENT_TIMEOUT = 5
+_THREAD_JOIN_TIMEOUT = 7
+msg = "thread wait must be larger than check wait"
+assert _THREAD_JOIN_TIMEOUT >= _CHECK_STOP_EVENT_TIMEOUT, msg
+
+
+@unique
+class Operation(Enum):
+    UPLOAD = "UPLOAD"
+    DELETE = "DELETE"
+    DELETE_TREE = "DELETE_TREE"
+
+
+def _get_requests_session() -> requests.Session:
+    retry = requests.adapters.Retry(
+        total=3,
+        backoff_factor=0.5,
+        status_forcelist=[429, 500, 502, 503, 504],
+        allowed_methods={"POST"},
+    )
+    adapter = requests.adapters.HTTPAdapter(max_retries=retry)
+    request_session = requests.Session()
+    request_session.mount("http://", adapter)
+    request_session.mount("https://", adapter)
+    return request_session
+
+
+@dataclass(frozen=True)
+class HTTPNotifyJob:
+    url: str
+    json: str
+
+
+def background_http_request(
+    session: requests.Session,
+    queue: "Queue[HTTPNotifyJob]",
+    stop_event: threading.Event,
+    stop_event_check_timeout: float,
+) -> None:
+    with session:
+        while not stop_event.is_set():
+            try:
+                notification = queue.get(timeout=stop_event_check_timeout)
+            except Empty:
+                notification = None
+
+            if notification is not None:
+                try:
+                    session.post(notification.url, notification.json)
+                except RequestException as e:
+                    LOG.warning(
+                        "[BackgroundHTTPNotifier] POST request to %r failed with %r, dropping event",
+                        notification.url,
+                        str(e),
+                    )
+
+                # The session is configured to do automatic retries, and that should handle most
+                # transients errors. Situations that takes longer than the configured retries will
+                # cause events to be dropped because we can't differentiate it from a non-transient
+                # errors.
+                #
+                # For non-transient errors the thread must kept running to consume from the queue,
+                # because there is no mechanism to stop producing to it. Hopefully the failure is
+                # fast enough that the queue won't grow faster than this threads consumes from it.
+                queue.task_done()
+
+
+def initialize_background_thread(
+    queue: Queue,
+    stop_event: threading.Event,
+    stop_event_check_timeout: float = _CHECK_STOP_EVENT_TIMEOUT,
+    session: requests.Session = None,
+) -> threading.Thread:
+    thread_session = session or _get_requests_session()
+    thread_args = (thread_session, queue, stop_event, stop_event_check_timeout)
+    thread = threading.Thread(
+        target=background_http_request,
+        args=thread_args,
+        name="BackgroundHTTPNotifierThread",
+        # The thread is a daemon thread to prevent it from blocking the shutdown. This would happen
+        # if the current notifier is not closed.
+        daemon=True,
+    )
+    thread.start()
+    return thread
+
+
+class BackgroundHTTPNotifier(Notifier):
+    def __init__(
+        self,
+        url: str,
+        stop_event_check_timeout: float = _CHECK_STOP_EVENT_TIMEOUT,
+        session: requests.Session = None,
+    ) -> None:
+        self._url = url
+        self._queue: "Queue[HTTPNotifyJob]" = Queue()
+        self._stop_event = threading.Event()
+        self._thread = initialize_background_thread(
+            self._queue,
+            self._stop_event,
+            stop_event_check_timeout,
+            session=session,
+        )
+
+    def __del__(self) -> None:
+        # Set the stop event on a best effort. If the event is not set the background thread leaks
+        # because it has a strong reference to itself.
+        self.close()
+
+    def close(self) -> None:
+        self._stop_event.set()
+        self._thread.join(_THREAD_JOIN_TIMEOUT)
+
+    def object_created(self, key: str, size: Optional[int]) -> None:
+        self._queue.put(
+            HTTPNotifyJob(
+                self._url,
+                json.dumps(
+                    {
+                        "op": Operation.UPLOAD.value,
+                        "key": key,
+                        "size": size,
+                        "last_modified": datetime.now(tz=timezone.utc).isoformat(),
+                    }
+                ),
+            )
+        )
+
+    def object_deleted(self, key: str) -> None:
+        self._queue.put(
+            HTTPNotifyJob(
+                self._url,
+                json.dumps(
+                    {
+                        "op": Operation.DELETE.value,
+                        "key": key,
+                    }
+                ),
+            )
+        )
+
+    def tree_deleted(self, key: str) -> None:
+        self._queue.put(
+            HTTPNotifyJob(
+                self._url,
+                json.dumps(
+                    {
+                        "op": Operation.DELETE_TREE.value,
+                        "key": key,
+                    }
+                ),
+            )
+        )
--- a/rohmu/notifier/interface.py
+++ b/rohmu/notifier/interface.py
@@ -1,0 +1,36 @@
+"""Copyright (c) 2022 Aiven, Helsinki, Finland. https://aiven.io/"""
+from abc import ABC, abstractmethod
+from typing import Optional
+
+
+class Notifier(ABC):
+    """This interface allows external code to be notified about object changes."""
+
+    @abstractmethod
+    def object_created(self, key: str, size: Optional[int]) -> None:
+        """Called when an object is created."""
+
+    @abstractmethod
+    def object_deleted(self, key: str) -> None:
+        """Called when an object is deleted.
+
+        Note: This may be called with each individual object as a side-effect
+        of `delete_tree`, for drivers that do not support the higher level
+        operation.
+        """
+
+    @abstractmethod
+    def tree_deleted(self, key: str) -> None:
+        """May be called when a tree is deleted.
+
+        Note: Not every driver supports this operation, for those objects will
+        be listed and deleted individually, in that case `object_deleted` is
+        called instead.
+        """
+
+    def object_copied(self, key: str, size: Optional[int]) -> None:
+        """Called when an object is copied."""
+        self.object_created(key=key, size=size)
+
+    def close(self) -> None:
+        """Method used to clean resources of the notifier, if any."""
--- a/rohmu/notifier/null.py
+++ b/rohmu/notifier/null.py
@@ -1,0 +1,19 @@
+"""Copyright (c) 2022 Aiven, Helsinki, Finland. https://aiven.io/"""
+from .interface import Notifier
+from typing import Optional
+
+
+class NullNotifier(Notifier):
+    """Empty implementation.
+
+    Used by default if configuration is missing to avoid None checks
+    """
+
+    def object_created(self, key: str, size: Optional[int]) -> None:
+        pass
+
+    def object_deleted(self, key: str) -> None:
+        pass
+
+    def tree_deleted(self, key: str) -> None:
+        pass
--- a/test/notifier/test_http.py
+++ b/test/notifier/test_http.py
@@ -1,0 +1,232 @@
+"""Copyright (c) 2022 Aiven, Helsinki, Finland. https://aiven.io/"""
+from contextlib import closing, contextmanager
+from datetime import datetime
+from http.server import BaseHTTPRequestHandler, HTTPServer
+from queue import Queue
+from rohmu.notifier.http import (
+    background_http_request,
+    BackgroundHTTPNotifier,
+    HTTPNotifyJob,
+    initialize_background_thread,
+    Operation,
+)
+from typing import Any, Iterator, List, Tuple
+
+import json
+import requests
+import threading
+import time
+
+
+class _TestSession:
+    def __init__(self, *args, **kwargs) -> None:  # pylint: disable=super-init-not-called,unused-argument
+        self.post_called: List[Tuple[Any, ...]] = list()
+
+    def post(self, *args, **kwargs) -> None:
+        self.post_called.append((args, kwargs))
+
+    def __enter__(self) -> None:
+        pass
+
+    def __exit__(self, type, value, traceback) -> None:  # pylint: disable=redefined-builtin
+        pass
+
+
+def _join_queue_with_timeout(queue: Queue, *, timeout: float, iteration: float = 0.1) -> None:
+    while queue.unfinished_tasks and timeout > 0.0:
+        time.sleep(iteration)
+        timeout -= iteration
+
+    assert not queue.unfinished_tasks, "queue is not empty after timeout"
+
+
+@contextmanager
+def _make_notifier(url: str) -> Iterator[BackgroundHTTPNotifier]:
+    # pylint: disable=protected-access
+
+    session = requests.Session()
+    notifier = BackgroundHTTPNotifier(
+        url=url,
+        stop_event_check_timeout=0.2,
+        session=session,
+    )
+
+    with closing(notifier):
+        yield notifier
+
+        assert notifier._thread.is_alive(), "background thread must not die during the test execution"
+
+    assert not notifier._thread.is_alive(), "background thread must be stopped once close is called"
+
+
+@contextmanager
+def _create_local_server() -> Iterator[Tuple[HTTPServer, List[Any]]]:
+    post_called: List[Any] = list()
+
+    class _TestServerRequestHandler(BaseHTTPRequestHandler):
+        def do_POST(self) -> None:
+            content_length = int(self.headers.get("Content-Length"))
+            data = self.rfile.read(content_length)
+            post_called.append((self.command, self.path, data))
+            self.send_response(200)
+            self.end_headers()
+
+    server = HTTPServer(("", 0), _TestServerRequestHandler)
+
+    try:
+        yield (server, post_called)
+    finally:
+        server.server_close()
+
+
+@contextmanager
+def _create_server_and_configured_notifier(
+    path: str,
+) -> Iterator[Tuple[BackgroundHTTPNotifier, HTTPServer, List[Any]]]:
+    """Creates a server and a notifier configured to communicated with that server.
+
+    Note: The server does not run in the background and should be driven by the
+    test with calls to `handle_request`.
+    """
+    with _create_local_server() as (server, post_called):
+        url = f"http://{server.server_name}:{server.server_port}{path}"
+
+        with _make_notifier(url=url) as notifier:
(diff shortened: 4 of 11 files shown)


rohmu/notifier/http.py
```diff
@@ -0,0 +1,93 @@
...
+        self._session.post(self._url, {"op": Operation.DELETE.value, "key": key})
+
+    def tree_deleted(self, key: str) -> None:
+        self._session.post(self._url, {"op": Operation.DELETE_TREE.value, "key": key})
+
+
+@dataclass
+class _HTTPNotifyJob:
+    url: str
+    data: Dict[str, Any]
```
hackaugusto, line 58: The type here is intentionally limited to what is used in this module, technically it could be:

```suggestion
    data: str | bytes | Mapping[str, Any] | Iterable[tuple[str, str | None]] | IO[Any]
```

But there is no benefit to the increased complexity

https://github.com/python/typeshed/blob/master/stubs/requests/requests/sessions.pyi#L48

rohmu/notifier/http.py
```diff
@@ -0,0 +1,93 @@
+"""Copyright (c) 2022 Aiven, Helsinki, Finland. https://aiven.io/"""
+from .interface import Notifier
+from datetime import datetime
+from enum import Enum, unique
+from queue import SimpleQueue
```
hackaugusto, line 5: Available since py3.7, we just dropped support for 3.6 so we can use it now.

rohmu/notifier/http.py
```diff
@@ -0,0 +1,93 @@
...
+
+class BackgroundHTTPNotifier(Notifier):
+    def __init__(self, url: str) -> None:
+        self._url = url
+        self._queue = SimpleQueue()
+        self._thread = threading.Thread(
+            target=_background_http_request,
+            args=(self._queue,),
+            name="BackgroundHTTPNotifierThread",
+            daemon=True,
```
hackaugusto, line 75: The publishing of the data is done on a best effort basis, so it is okay for the queue to have a few items left over when the interpreter exits.

The alternative would be to introduce synchronization for cleanly shutting down the background threads, which would introduce more changes to the codebase. If we ever want to make a stronger guarantee we can add it later.

rohmu/notifier/http.py
```diff
@@ -0,0 +1,98 @@
+"""Copyright (c) 2022 Aiven, Helsinki, Finland. https://aiven.io/"""
```
hackaugusto, line 1: This module is basically only doing IO. I could write tests for it, but they would basically endup testing the mock implementation instead because there is not much logic besides logging / http requests.

rohmu/notifier/interface.py
```diff
@@ -0,0 +1,32 @@
+"""Copyright (c) 2022 Aiven, Helsinki, Finland. https://aiven.io/"""
+from abc import ABC, abstractmethod
+from typing import Optional
+
+
+class Notifier(ABC):
+    """This interface allows external code to be notified about object changes."""
+
+    @abstractmethod
+    def object_created(self, key: str, size: Optional[int]) -> None:
```
ngilles-aiven, line 10: nit: missing docstring on this one compared to the others?

```suggestion
    def object_created(self, key: str, size: Optional[int]) -> None:
    """Called when and object is created."""
```

rohmu/notifier/null.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+
+    def object_created(self, key: str, size: Optional[int]) -> None:
+        pass
+
+    def object_deleted(self, key: str) -> None:
+        pass
+
+    def tree_deleted(self, key: str) -> None:
+        pass
```
ngilles-aiven, line 19: Should we create a (singleton) instance of this one that could be used a default value for notifier params (avoiding `Optional`s)

rohmu/notifier/http.py
ngilles-aiven: Yep, leaving it to what data we want to transit is quite reasonable IMHO

rohmu/notifier/http.py
```diff
@@ -0,0 +1,98 @@
...
+        self._session.post(self._url, {"op": Operation.DELETE_TREE.value, "key": key})
+
+
+@dataclass
+class _HTTPNotifyJob:
+    url: str
+    data: Dict[str, Any]
+
+
+def _background_http_request(queue: "SimpleQueue[_HTTPNotifyJob]") -> None:
```
ngilles-aiven, line 62: Q: is there a specific reason this a standalone function, and not a method on `BackgroundHTTPNotifier`? almost feels like it was going to be to implement a single thread(pool?) that could handle multiple notifiers?

ngilles-aiven (commented):
LGTM, with curiousity question and a nit.

rohmu/notifier/http.py
```diff
@@ -0,0 +1,98 @@
...
+    with _get_requests_session() as session:
+        while True:
+            notification = queue.get()
+            session.post(notification.url, notification.data)
```
sjamgade, line 66: The notification is dropped if the `post` fails
Also,  it will also kill the thread?

ngilles-aiven: That the notification is dropped is ok, it's a notification we're doing best effort here (sessions is configured to retry 3 times), but we should probably just log it and you are right @sjamgade not kill the thread.

rohmu/notifier/http.py
sjamgade: one failure in post request will cause all further requests to fail. A threadpool make sure all threads are cleaned when the pool object goes out of scope. currently there is a `q.get` which is blocking. so unless  something asks that `q.get` to return the thread is going to blocked

rohmu/notifier/http.py
```diff
@@ -0,0 +1,98 @@
...
+                "size": size,
+                "last_modified": datetime.now(),
+            },
+        )
+
+    def object_deleted(self, key: str) -> None:
+        self._session.post(self._url, {"op": Operation.DELETE.value, "key": key})
+
+    def tree_deleted(self, key: str) -> None:
+        self._session.post(self._url, {"op": Operation.DELETE_TREE.value, "key": key})
```
ngilles-aiven, line 53: Building on Sumit's comments would these also break rhome uploads if there is an error, as they're called inline in the main thread?

rohmu/notifier/http.py
sjamgade: dropping is fine, but something is also not difficult given that we are in a background thread
```suggestion
            try:
                 session.post(notification.url, notification.data)
            except SomeException: 
                     log_the_exception
                     time.sleep(5)
                     queue.put(notification)
```

rohmu/notifier/http.py
hackaugusto: > so unless something asks that q.get to return the thread is going to blocked

That was intentional, it is simpler this way. This instead doesn't block the python process on exit because that thread never leaves.

> one failure in post request will cause all further requests to fail.

Will add exception handling.

sjamgade: > This instead doesn't block the python process on exit because that thread never leaves.

can you say that again ? do you mean **does** block ? so restarting *hoards and astacus will get blocked, unless  ?
The interpreter does get blocked.

```

In [1]: from threading import Thread

In [2]: from queue import Queue

In [3]: def wp(q):
   ...:     while True:
   ...:         o = q.get()
   ...:         print(o)
   ...:

In [4]: t = Thread(target=wp, args=(Queue(),))

In [5]: t.start()

In [6]:
Do you really want to exit ([y]/n)? y


^CException ignored in: <module 'threading' from '/usr/lib64/python3.10/threading.py'>
Traceback (most recent call last):
  File "/usr/lib64/python3.10/threading.py", line 1558, in _shutdown
    lock.acquire()
KeyboardInterrupt:
➜  aiven-core git:(13653b8e4a)
```

rohmu/notifier/interface.py
hackaugusto: Added as per the request. Note that I don't think this is improving things, since the comment says the same thing that is expressed by the method name. The other methods had docstrings because it required a note to express its behavior.

rohmu/notifier/null.py
hackaugusto: There is no need for a singleton, I initialized the value on the `BaseTransfer` class:

https://github.com/aiven/rohmu/pull/67/files#diff-19b8216f21f883fff1a37d854110c70b40006e6be762e8251178b8f654247285R30

rohmu/notifier/http.py
hackaugusto: Maybe I should be more explicit, the line this comment is added to is important, namely you **need** the flag `daemon=True` for the behavior I mentioned

sjamgade: yup, missed the `daemon` flag entirely

rohmu/notifier/http.py
```diff
@@ -0,0 +1,132 @@
...
+    )
+    adapter = requests.adapters.HTTPAdapter(max_retries=retry)
+    request_session = requests.Session()
+    request_session.mount("http://", adapter)
+    return request_session
+
+
+class HTTPNotifier(Notifier):
+    def __init__(self, url: str) -> None:
+        self._session = _get_requests_session()
```
kmichel-aiven, line 39: This leaks a `requests.Session` every time the object is instantiated.

rohmu/notifier/http.py
```diff
@@ -0,0 +1,132 @@
...
+def _get_requests_session() -> requests.Session:
+    retry = requests.adapters.Retry(
+        total=3,
+        backoff_factor=0.5,
+        status_forcelist=[429, 500, 502, 503, 504],
+        method_whitelist={"POST"},
+    )
+    adapter = requests.adapters.HTTPAdapter(max_retries=retry)
+    request_session = requests.Session()
+    request_session.mount("http://", adapter)
```
kmichel-aiven, line 33: Why not including https ?

rohmu/notifier/http.py
```diff
@@ -0,0 +1,132 @@
...
+        except RequestException as e:
+            LOG.warning(
+                "[HTTPNotifier] POST request to %r failed with %r, dropping event",
+                self._url,
+                str(e),
+            )
+
+
+@dataclass
+class _HTTPNotifyJob:
```
kmichel-aiven, line 75: This should be `frozen=True` and the data field should be `Mapping` instead of `Dict`, especially since it crosses a threading boundary.

Actually, we shouldn't probably even have a generic `data`, making a dataclass but putting almost everything in a dict in a single field of the dataclass isn't great: serialization to HTTP-friendly but untyped format should be done as necessary when leaving the system, not before.

rohmu/notifier/http.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def object_created(self, key: str, size: Optional[int]) -> None:
+        self._queue.put(
+            _HTTPNotifyJob(
+                self._url,
+                {
+                    "op": Operation.UPLOAD.value,
+                    "key": key,
+                    "size": size,
+                    "last_modified": datetime.now(),
```
kmichel-aiven, line 123: This should probably be `utcnow`, unless we want interesting bugs during summer time :)

rohmu/notifier/http.py
```diff
@@ -0,0 +1,132 @@
...
+                    notification.url,
+                    str(e),
+                )
+
+
+class BackgroundHTTPNotifier(Notifier):
+    def __init__(self, url: str) -> None:
+        self._url = url
+        self._queue: "SimpleQueue[_HTTPNotifyJob]" = SimpleQueue()
+        self._thread = threading.Thread(
```
kmichel-aiven, line 107: This leaks a thread every time the object is instantiated.

rohmu/notifier/http.py
```diff
@@ -0,0 +1,132 @@
...
+        status_forcelist=[429, 500, 502, 503, 504],
+        method_whitelist={"POST"},
+    )
+    adapter = requests.adapters.HTTPAdapter(max_retries=retry)
+    request_session = requests.Session()
+    request_session.mount("http://", adapter)
+    return request_session
+
+
+class HTTPNotifier(Notifier):
```
kmichel-aiven, line 37: I don't think this implementation should exist at all, there is already the `BackgroundHTTPNotifier` and this one creates unwanted behavior since it blocks on the notification query.

rohmu/notifier/interface.py
```diff
@@ -0,0 +1,33 @@
+"""Copyright (c) 2022 Aiven, Helsinki, Finland. https://aiven.io/"""
+from abc import ABC, abstractmethod
+from typing import Optional
+
+
+class Notifier(ABC):
```
kmichel-aiven, line 6: Given the resource leaks in `(Backgrond)HTTPNotifier`, this interface should probably include at least a `stop` function to cleanup resources.

kmichel-aiven (dismissed):
Mostly some small changes, the Notifier API needs some `stop` function though, we can't have a API that by design requires leaking resources.

And all these classes require tests (the `responses` library would probably be a good fit to test `requests` things).

rohmu/notifier/interface.py
ngilles-aiven: of course I managed to slip in a typo in half a sentence :'(  `and` -> `an` sorry :-/

Yeah in this case the comment doesn't look like it provides much benefit, except maybe consistency, but down the line if we generate source docs it will be a tad better.

rohmu/notifier/http.py
ngilles-aiven: That's true, I've usually designed all my threading stuff by having classes inheriting from `Thread` and have my class basically be the the isolation level/context for the thread. Same result different approach

rohmu/notifier/http.py
ngilles-aiven: Maybe we should explicitly add the explanation as code comment?

rohmu/notifier/interface.py
ngilles-aiven: Candiate for becoming a context manager too at this point?

rohmu/notifier/http.py
hackaugusto: Why would there be more than one instance of such an object?

rohmu/notifier/http.py
hackaugusto: if both classes are exposed and their behavior is documented, why would you claim the user configuring this is unwanted behavior?

anyways, will remove since we don't need it, and it is easy enough to add if we ever need it

rohmu/notifier/http.py
ngilles-aiven: I suppose anyone using rhomu in their loop could potentially create a new notifier on each iteration (even though they technically wouldn't have to

rohmu/notifier/interface.py
ngilles-aiven: Although I'm thinking it's nothing inherent in the concept of a `Notifier` that really make a "resource" thingy... maybe just the `(Background)HTTPNotifier` class should implement that (and be accordingly documented) ?

rohmu/notifier/http.py
hackaugusto: > That's true, I've usually designed all my threading stuff by having classes inheriting from Thread and have my class basically be the the isolation level/context for the thread. Same result different approach

I would agree if that was in a different class, and not on `BackgroundHTTPNotifier`. If you meant something like, `Thread(target=my_logic)` vs `class MyLogic(Thread)`, I guess it is only a matter of taste.

rohmu/notifier/http.py
hackaugusto: > This should be frozen=True and the data field should be Mapping instead of Dict, especially since it crosses a threading boundary.

good point, done.

> Actually, we shouldn't probably even have a generic data, making a dataclass but putting almost everything in a dict in a single field of the dataclass isn't great: serialization to HTTP-friendly but untyped format should be done as necessary when leaving the system, not before.

I don't understand what you're trying to say. Are you proposing implementing a separate dataclass for each message type with (de)serialization? I would prefer to not do that since that would only increase code size without any benefits.

rohmu/notifier/http.py
hackaugusto: changed to a time-aware datatime with UTC.

rohmu/notifier/http.py
hackaugusto: added `https://`

rohmu/notifier/http.py
hackaugusto: Okay, but if people are using things like that we most likely are leaking every transfer object, the `BaseTransfer` has not `stop` or `close` method

hackaugusto: Is that code relying on garbage collection to release resources?

ngilles-aiven: actually `Session might` be holding connection pools/threads too?

hackaugusto:
> And all these classes require tests (the responses library would probably be a good fit to test requests things).

I intentionally didn't write tests.

Do you really want it for everything, including this?

https://github.com/aiven/rohmu/blob/d125642d7dcc480d2922e0b44be0f6d025dd51a5/rohmu/notifier/null.py#L6-L19

i will add tests to the background notifier class

rohmu/notifier/http.py
hackaugusto: well, if the `BaseTransfer` is relying on the GC, since that is used to handle cycles, this should be fine too (random doc [about cycles and the gc](https://docs.python.org/3/c-api/gcsupport.html))

Sample script:

```python
from dataclasses import dataclass, field


@dataclass
class Child:
    parent: "Parent"

    def __del__(self) -> None:
        print("Child is gone")


@dataclass
class Parent:
    child: Child = field(init=False)

    def __post_init__(self) -> None:
        child = Child(self)
        self.child = child

    def __del__(self) -> None:
        print("Parent is gone")


Parent()
```

Just to be clear, I'm by no means stating this is good design. I am, however, trying not rewrite the whole thing.

hackaugusto: Hummm, actually let me check something. Because we are going over a thread boundary tehre.

rohmu/notifier/http.py
kmichel-aiven: Yes, I'm suggesting creating separate dataclass type for different message types :) Serialization can stay very simple and in a parent class though. It's mainly to avoid exchanging bundles of untyped dicts between parts of the code.

We've seen too many bugs caused by use of dictionaries that could have been caught by static type checks: missing key, wrong type of value, surprising None... Not having this type of bugs anymore would be the main benefit, the cost is verbosity but it's a lesser cost than production bug. (We already have some dicts in other place, but we want to have less of them over time).

rohmu/notifier/interface.py
kmichel-aiven: Would be annoying if the caller has to sniff the exact `Notifier` subclass(es) generated by `get_transfer` to decide if it should call `close`. It's better to have `close` part of the interface even if it sometimes does nothing, and it will give a subtle hint to the future implementer of another `Notifier` class that it's a good idea to think about cleanup :slightly_smiling_face:

rohmu/notifier/http.py
kmichel-aiven: Good point, we'll have to make sure `BaseTransfer` can also be closed (and it might be the one responsible for cleaning up its notifiers).

rohmu/notifier/interface.py
hackaugusto: added `close` to the interface.

rohmu/notifier/http.py
hackaugusto: I agree that passing dicts around is bad in some cases, however I don't think having a model here would be benefitial, unless the type is exported as a model to be used by the consumer of the notifier data. Is that what you mean? You want these to be implemented with something like `pydantic` models?

kmichel-aiven: Ok, since it does not leave BackgroundHTTPNotifier we can live with the dict.

(I didn't mean pydantic, just dataclass, the type coercion of pydantic is not very helpful to avoid typing bugs :smile:)

sjamgade (approved):
LGTM in its current form

kmichel-aiven:
> > And all these classes require tests (the responses library would probably be a good fit to test requests things).
> 
> I intentionally didn't write tests.
> 
> Do you really want it for everything, including this?
> 
> https://github.com/aiven/rohmu/blob/d125642d7dcc480d2922e0b44be0f6d025dd51a5/rohmu/notifier/null.py#L6-L19
> 
> i will add tests to the background notifier class

Yes, we really want high test coverage on this project (one part is testing all "logic" code, the other is setting up various cloud accounts to add integration tests, this is in progress).
There is a history of production issues to motivate this, since it's backup-related we really don't want more breakage.
This includes even seemingly trivial code: even if a piece of code performs some non-critical operation, its very easy for one small failure to bubble up and break unrelated critical operations.

hackaugusto:
Those methods are literally empty, I'm not sure what you wanted, but I added a test that only calls those.

kmichel-aiven (dismissed):

test/notifier/test_http.py
```diff
@@ -0,0 +1,166 @@
...
+        self.post_called.append((args, kwargs))
+
+    def __enter__(self) -> None:
+        pass
+
+    def __exit__(self, type, value, traceback) -> None:  # pylint: disable=redefined-builtin
+        pass
+
+
+def test_background_http_request() -> None:
```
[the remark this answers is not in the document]
hackaugusto: simple queue doesn't have an interface to join on results:

https://docs.python.org/3/library/queue.html#simplequeue-objects

and using a full fledged queue is incompatible (it requires calls to `task_done` which is not available on `SimpleQueue`)

https://docs.python.org/3/library/queue.html#queue.Queue.join

I tought about replacing the implementation, but then the test would block forever and not give an error, which IMO is worse than getting a failure with an assert. Unfortunally if we have to test IO, we need timeouts.

test/notifier/test_http.py
```diff
@@ -0,0 +1,166 @@
...
+        stop_event_check_timeout=stop_event_check_timeout,
+    )
+    assert thread.is_alive(), "freshly created thread must be healthy and running"
+    assert thread.daemon is True, "background thread must be daemon so that it won't prevent main thread from exiting"
+    stop_event.set()
+    time.sleep(stop_event_check_timeout + 0.01)  # wait for the thread to check the event is set
+    assert not thread.is_alive(), "thread must exit after stop_event is set"
+
+
+def test_BackgroundHTTPNotifier() -> None:
```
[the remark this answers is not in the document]
kmichel-aiven: You can use a module-scope fixture for the test server.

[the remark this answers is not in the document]
hackaugusto: found a solution

test/notifier/test_http.py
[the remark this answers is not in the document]
hackaugusto: Both of the things you suggest are implemented with `time.sleep`, as per you sample. The only thing that changes is how the error is manifested.

Will change to Queue

kmichel-aiven (approved):

hackaugusto:
thanks @kmichel-aiven 🙇

PR #67, opened 2022-09-09
Integrate notifier with drivers

hackaugusto:
~Merge/Review after https://github.com/aiven/rohmu/pull/66~

Add support for the notifier to each driver

PR #68, opened 2022-09-16
Version 1.0.5

heikju:
Bump version number to 1.0.5 for a new release

Resolves: #xxxxx

# Why this way

### DIFF
--- a/rohmu/version.py
+++ b/rohmu/version.py
@@ -1,1 +1,1 @@
-VERSION = "1.0.4"
+VERSION = "1.0.5"


kmichel-aiven (changes requested):
We should not release until https://github.com/aiven/rohmu/pull/67 is reverted or fixed,
this PR broke the Azure, Swift and SFTP storage engines.

heikju:
Reverting that would revert https://github.com/aiven/rohmu/pull/67/commits/f164f479347df12f20e57582e8f5ba8cc5c119a6, so it seems it reverting in its entirety would mean the notifications couldn't actually be enabled.

Any hints for how I could go about reproducing the failures with those storage engines to be able to fix this?

kmichel-aiven:
> Reverting that would revert [f164f47](https://github.com/aiven/rohmu/commit/f164f479347df12f20e57582e8f5ba8cc5c119a6), so it seems it reverting in its entirety would mean the notifications couldn't actually be enabled.
> 
> Any hints for how I could go about reproducing the failures with those storage engines to be able to fix this?

We do not have the setup for automated integration tests currently, it's in progress.

In the meantime, you can either write mocks for all storage types to be able to unit-test this or do a manual test with each of them using custom code or any project that uses rohmu (pghoard, myhoard, astacus...) and report on the manual tests results.

heikju:
I think I'll go the unittest way, starting with the storage modules reportedly broken. At least the calls to `os.path.getsize()` with what seem to be non-pathname args seem dubious, so let's see if I can repro this with just tests.

hackaugusto:
> We should not release until https://github.com/aiven/rohmu/pull/67 is reverted or fixed,
> this PR broke the Azure, Swift and SFTP storage engines.

@kmichel-aiven do you have some reference to the bugs? I can fix it I have some sample code / tests.

alanfranz:
> > We should not release until #67 is reverted or fixed,
> > this PR broke the Azure, Swift and SFTP storage engines.
> 
> @kmichel-aiven do you have some reference to the bugs? I can fix it I have some sample code / tests.

@hackaugusto I'm following this issue because I've got a PR that I'd like to have applied on rohmu and it's now blocked waiting for this problem to resolve.

I tried with SFTP since it was the easiest for me, and I got an error:

```
s = SFTPTransfer("microserver.home.franzoni.net", 22, "alan", password="XXXX", prefix="/tmp")
s.store_file_from_memory(key="something", memstring=b"aaaa", metadata={"a":"b"})
```

```(venv) alan.franzoni@toolbox:~/aiven/rohmu(main)$ python -m rohmu.object_storage.sftp.py 
Traceback (most recent call last):
  File "/usr/lib64/python3.10/runpy.py", line 187, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/usr/lib64/python3.10/runpy.py", line 110, in _get_module_details
    __import__(pkg_name)
  File "/home/alan.franzoni/aiven/rohmu/rohmu/object_storage/sftp.py", line 242, in <module>
    s.store_file_from_memory(key="something", memstring=b"aaaa", metadata={"a":"b"})
  File "/home/alan.franzoni/aiven/rohmu/rohmu/object_storage/sftp.py", line 187, in store_file_from_memory
    self.store_file_object(key=key, fd=bio, cache_control=cache_control, metadata=metadata, mimetype=mimetype)
  File "/home/alan.franzoni/aiven/rohmu/rohmu/object_storage/sftp.py", line 213, in store_file_object
    self.notifier.object_created(key=key, size=os.path.getsize(fd))
  File "/usr/lib64/python3.10/genericpath.py", line 50, in getsize
    return os.stat(filename).st_size
TypeError: stat: path should be string, bytes, os.PathLike or integer, not BytesIO
```

This works properly at 6def954ac01746f111c65652c83e3750bcfb45ae 

Two things:
* I think you shouldn't ask for tests; instead, tests (in some form) should be provided with the code. the absence of proof is not proof of absence :-)
* I think that a few of those errors could be catched by just adding typing information and running mypy.

One architectural consideration since I'm at it: I think that this approach that puts notification code everywhere makes it harder to maintain, test and debug the code. As an Old Java Guy I'd suggest using the decorator pattern; you could define a class like

```class NotifyingTransfer(BaseTransfer):
    def __init__(self, inner: BaseTransfer, notifier: Notifier):
        ...
```

that exposes the very same interface as a basetransfer, delegates to ```inner``` (which is a concrete BaseTransfer implementation) and notifies before/after methods are called. This would yield no need to change each & every basetransfer subclass (i didn't fully check the code, maybe there's something that actually needs to go inside each implementation). 

This way, in order to test the code you add, it would be enough to create a StubSomethingTransfer (or maybe just use a LocalTransfer) and you'd be done. No need for cloud-related integration tests.

hackaugusto:
> I think you shouldn't ask for tests; instead, tests (in some form) should be provided with the code. the absence of proof is not proof of absence :-)

You're correct that `absence of proof is not proof of absence`, but this felt like bug report and usually I expect a MRE with them.

I based the code on an existing PR https://github.com/aiven/rohmu/pull/59/files#diff-d478aeda90fd0457fe984ceede41b062982a3db70ce50e3cc9ab78ed5ff44ffbR386 and assumed the code was tested.

> [..] the decorator pattern; you could define a class like

That was considered and dropped, please refer to: https://github.com/aiven/rohmu/pull/59#issuecomment-1237996851 , there are also an extensive discussion on our internal chat.

I believe @heikju is working on tests and a fix.

heikju:
https://github.com/aiven/rohmu/pull/70

heikju:
I'll close this, we can try again after we've gotten rohmu fixes in.

PR #69, opened 2022-09-19
Make sure get_contents_to_file() calls are atomic

alanfranz:
Make sure that calls to get_contents_to_file() are actually atomic

Calls to get_contents_to_file() weren't necessarily atomic - destination file
could be visible before it was completely written, or it could be left half-written
on the filesystem if the process crashed, or it could leave other temporary files
that nothing would cleanup in the same directory as target file.

This was related to some bugs that can be seen in pghoard.

We fix all these issues through an anonymous temporary file approach; details
in commit messages.

Moved over from https://github.com/aiven/rohmu/pull/60 since I changed GH username and the PR wasn't mine anymore.

### DIFF
--- a/test/test_atomic_opener.py
+++ b/test/test_atomic_opener.py
@@ -1,0 +1,145 @@
+from pathlib import Path
+from rohmu.atomic_opener import atomic_opener
+
+import errno
+import os
+import pytest
+import time
+
+
+def _verify_file_not_created_and_dir_not_polluted(output_file: Path):
+    assert os.listdir(output_file.parent) == []
+    assert not output_file.exists()
+
+
+def test_error_thrown_if_final_path_parent_doesnt_exist(tmp_path: Path):
+    with pytest.raises(IOError):
+        with atomic_opener(tmp_path / "nonexistingdir" / "final_path", mode="w") as f:
+            pass
+
+
+def test_error_mode_doesnt_contain_write(tmp_path: Path):
+    with pytest.raises(ValueError):
+        with atomic_opener(tmp_path, mode="r") as f:  # type: ignore
+            pass
+
+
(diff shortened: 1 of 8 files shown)


test/test_atomic_opener.py
```diff
@@ -0,0 +1,145 @@
+from pathlib import Path
+from rohmu.atomic_opener import atomic_opener
+
+import errno
+import os
+import pytest
+import time
+
+
+def _verify_file_not_created_and_dir_not_polluted(output_file: Path):
```
kmichel-aiven, line 10: return type hints are missing on all functions

kmichel-aiven (approved):

test/test_atomic_opener.py
alanfranz: I can fix that.

Maybe I'm misremembering, though, I need a mypy cheatsheet: if there's at least one typed argument, mypy should perform type checking nevertheless, and treat return type as None by default, isn't this the default behaviour?

kmichel-aiven: mypy will usually type check (depending on its settings, which we need to recheck on this repo).

However the default type, if untyped, is not `None`, it's `Any`. So if you have something like:

```python
def check_is_odd(x: int):
    return x % 2

def test_something():
   x = something_give_me_a_value()
   check_is_odd(x)
```

The caller and the callee did not agree on whether it's supposed to be a boolean function or an assert-like function,
By not typing the return type, you miss an opportunity to be warned about it.

(This is a simplified but seen-before bug :) )

kmichel-aiven: But we really need to fix the mypy settings, so that everybody writes the full function prototype before review-time :)

PR #70, opened 2022-09-20
notifier: fix notifications for some storage types, add tests

heikju:
For Azure, Swift and SFTP object storages, the attempt to get object size for completion notifications was incorrect, calling `os.path.getsize()` with a file-like object, instead of a path.

Given that the object to store may not be a file on disk, that can be seeked or `stat()`ed, we need some other way to obtain the number of bytes.

For these storages, the store call doesn't seem to return the object size in a convenient way, so make use of the progress callback we implement for object storage backends. That gives us the number of bytes transferred, that we can then pass on to notifier.

For tests, start with these three backends that seemed to have been broken by #67. Some libraries used by them are not in our requirements.dev.txt, so settle for mocking those imports.

Looks to fix a few storage backends broken by #67, and adds new tests

Resolves: #xxxxx

# Why this way

### DIFF
--- a/rohmu/object_storage/azure.py
+++ b/rohmu/object_storage/azure.py
@@ -345,11 +345,14 @@ class AzureTransfer(BaseTransfer):
         content_settings = None
         if mimetype:
             content_settings = ContentSettings(content_type=mimetype)
+        notify_size = [0]
 
         def progress_callback(pipeline_response):
             bytes_sent = pipeline_response.context["upload_stream_current"]
-            if upload_progress_fn and bytes_sent:
-                upload_progress_fn(bytes_sent)
+            if bytes_sent:
+                notify_size[0] = bytes_sent
+                if upload_progress_fn:
+                    upload_progress_fn(bytes_sent)
 
         # Azure _BlobChunkUploader calls `tell()` on the stream even though it doesn't use the result.
         # We expect the input stream not to support `tell()` so use dummy implementation for it
@@ -367,7 +370,7 @@ class AzureTransfer(BaseTransfer):
                 raw_response_hook=progress_callback,
                 overwrite=True,
             )
-            self.notifier.object_created(key=key, size=os.path.getsize(fd))
+            self.notifier.object_created(key=key, size=notify_size[0])
         finally:
             if not seekable:
                 if original_tell is not None:
--- a/rohmu/object_storage/sftp.py
+++ b/rohmu/object_storage/sftp.py
@@ -11,6 +11,7 @@ from ..notifier.interface import Notifier
 from .base import BaseTransfer, IterKeyItem, KEY_TYPE_OBJECT, KEY_TYPE_PREFIX
 from io import BytesIO, StringIO
 from stat import S_ISDIR
+from typing import cast
 
 import datetime
 import json
@@ -54,7 +55,7 @@ class SFTPTransfer(BaseTransfer):
         else:  # password must be defined due to previous check above
             transport.connect(username=self.username, password=self.password)
 
-        self.client = paramiko.SFTPClient.from_transport(transport)
+        self.client = cast(paramiko.SFTPClient, paramiko.SFTPClient.from_transport(transport))
 
         self.log.debug("SFTPTransfer initialized")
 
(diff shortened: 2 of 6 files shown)


rohmu/object_storage/swift.py
```diff
@@ -297,17 +297,24 @@ def copy_file(self, *, source_key, destination_key, metadata=None, **_kwargs):
...
             cache_control=cache_control,
             metadata=metadata,
             mimetype=mimetype,
-            upload_progress_fn=upload_progress_fn,
+            upload_progress_fn=upload_progress_fn_wrapper,
             multipart=True,
             content_length=metadata.get("Content-Length"),
         )
-        self.notifier.object_created(key=key, size=os.path.getsize(fd))
+        self.notifier.object_created(key=key, size=notify_size[0])
```
kmichel-aiven, line 317: Since the size will only be available when metadata.get("Content-Length") is available, why not using that directly instead of wrapping the callback just to get back the same value ?

rohmu/object_storage/sftp.py
```diff
@@ -192,15 +192,17 @@ def store_file_from_memory(self, key, memstring, metadata=None, cache_control=No
...
 
     # pylint: disable=unused-argument
     def store_file_object(self, key, fd, *, cache_control=None, metadata=None, mimetype=None, upload_progress_fn=None):
         target_path = self.format_key_for_backend(key.strip("/"))
+        notify_size = [0]
 
         self.log.debug("Store path: %r", target_path)
 
         def wrapper_upload_progress_fn(bytes_written, total_bytes):  # pylint: disable=unused-argument
+            notify_size[0] = bytes_written
```
kmichel-aiven, line 205: This should accumulate the written bytes, not overwrite the previous value.

rohmu/object_storage/azure.py
```diff
@@ -345,11 +345,14 @@ def store_file_object(self, key, fd, *, cache_control=None, metadata=None, mimet
...
         if mimetype:
             content_settings = ContentSettings(content_type=mimetype)
+        notify_size = [0]
 
         def progress_callback(pipeline_response):
             bytes_sent = pipeline_response.context["upload_stream_current"]
-            if upload_progress_fn and bytes_sent:
-                upload_progress_fn(bytes_sent)
+            if bytes_sent:
+                notify_size[0] = bytes_sent
```
kmichel-aiven, line 353: This should accumulate the written bytes, not overwrite the previous value.

rohmu/object_storage/swift.py
```diff
@@ -297,17 +297,24 @@ def copy_file(self, *, source_key, destination_key, metadata=None, **_kwargs):
 
     def store_file_object(self, key, fd, *, cache_control=None, metadata=None, mimetype=None, upload_progress_fn=None):
         metadata = metadata or {}
+        notify_size = [0]
+
+        def upload_progress_fn_wrapper(bytes_written):
+            notify_size[0] = bytes_written
```
kmichel-aiven, line 303: If this is kept (cf. comment below), it should accumulate the written bytes, not overwrite the previous value.

test/test_object_storage_sftp.py
```diff
@@ -0,0 +1,55 @@
...
+            notifier=notifier,
+        )
+        test_data = b"test-data"
+        with NamedTemporaryFile() as tmpfile:
+            tmpfile.write(test_data)
+            tmpfile.flush()
+            transfer.store_file_from_disk(key="test_key1", filepath=tmpfile.name)
+
+        client.putfo.assert_called()
+        notifier.object_created.assert_any_call(key="test_key1", size=len(test_data))
```
alanfranz, line 27: I think this assertion is wrong. I'd assume (but you can prove me wrong) that we get 1 notification for each 1 object created, so this should assert that method 'object_created' is called once.

But I see that sftp store_file_from_disk calls store_file_object and both of them perform a notification, so there could be a bug there.

alanfranz (dismissed):
I spotted a problem, I can't finish the whole review though right now.

test/test_object_storage_azure.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@pytest.fixture(scope="module", name="mock_azure_module")
+def fixture_mock_azure_module():
+    get_blob_client_mock = MagicMock()
+    blob_client = MagicMock(get_blob_client=get_blob_client_mock)
+    service_client = MagicMock(from_connection_string=MagicMock(return_value=blob_client))
+    sys.modules["azure.common"] = MagicMock()
+    sys.modules["azure.core.exceptions"] = MagicMock()
+    sys.modules["azure.storage.blob"] = MagicMock(BlobServiceClient=service_client)
```
kmichel-aiven, line 17: This manually alters global variables in a remote module and does not revert it at the end of the test.
This should use something like [mock.patch](https://docs.python.org/dev/library/unittest.mock.html#unittest.mock.patch).

test/test_object_storage_azure.py
```diff
@@ -0,0 +1,77 @@
+"""Copyright (c) 2022 Aiven, Helsinki, Finland. https://aiven.io/"""
```
kmichel-aiven, line 1: Please add parameter and return type hints on all new code.

test/test_object_storage_swift.py
```diff
@@ -0,0 +1,56 @@
...
+from tempfile import NamedTemporaryFile
+from unittest.mock import MagicMock
+
+import pytest
+import sys
+
+
+@pytest.fixture(scope="module", name="swift_module")
+def fixture_swift_module():
+    sys.modules["swiftclient"] = MagicMock()
```
kmichel-aiven, line 12: Same as azure, this should use `mock.patch`

test/test_object_storage_swift.py
```diff
@@ -0,0 +1,56 @@
...
+        user="testuser",
+        key="testkey",
+        container_name="test_container",
+        auth_url="http://auth.example.com",
+        notifier=notifier,
+    )
+    test_data = b"test-data"
+    file_object = BytesIO(test_data)
+
+    transfer.store_file_object(key="test_key2", fd=file_object, metadata={"Content-Length": len(test_data)})
```
kmichel-aiven, line 53: The case without `Content-Length` should be either fixed to work like the other or be tested, separately.

kmichel-aiven (changes requested):
We found a bug in these three storages by reading their code, but all the storages modified in #67 should be tested.

test/test_object_storage_sftp.py
heikju: Yes, we can use `assert_called_once_with()` everywhere to be more strict, no reason really to not do so.

test/test_object_storage_azure.py
heikju: `patch.dict()` seems like the simplest way to do the trick.

rohmu/object_storage/sftp.py
heikju: Really? Paramiko docs say that the first arg is the number of bytes sent so far, and callback implementation here https://github.com/paramiko/paramiko/blob/main/paramiko/sftp_client.py#L675 seems to agree with that, already doing the accumulation of individual writes.

rohmu/object_storage/azure.py
heikju: It seems to me that the Azure client already accumulates this internally, though it turns out to be harder to find documentation about this, I wonder how recent the version we're using is. In any case, see line 301 of `/usr/lib/python3.10/site-packages/azure/storage/blob/_shared/policies.py`

rohmu/object_storage/swift.py
heikju: It seems to me that this is called with the value content-length minus number of bytes not yet transferred, unless I misunderstand the logic in `_store_file_contents()`

kmichel-aiven: Same as the previous two :+1:

heikju:
I hope I've addressed the comments to a sufficient degree. In the interest of making reviewing easier, I'd like to leave tests for the remaining storage backends to a follow-up PR.

heikju:
Updated to make type hints acceptable for Python 3.7

hackaugusto (approved):
LGTM, only had a small comment.

PR #71, opened 2022-09-21
swift: fix iter_key on empty string

rominf:

Listing all keys on root level for Swift requires passing `""` as a key.
    
Currently, in swift `iter_key` converts empty key (`""`) to `"/"`, which makes Swift return empty list of results even when there should be results.
    
This change fixes the issue.

# Why this way

Because it is the simpliest way to fix this issue.

### DIFF
--- a/rohmu/object_storage/swift.py
+++ b/rohmu/object_storage/swift.py
@@ -131,7 +131,7 @@ class SwiftTransfer(BaseTransfer):
         return metadata
 
     def iter_key(self, key, *, with_metadata=True, deep=False, include_key=False):
-        path = self.format_key_for_backend(key, trailing_slash=not include_key)
+        path = self.format_key_for_backend(key, remove_slash_prefix=True, trailing_slash=not include_key)
         self.log.debug("Listing path %r", path)
         if not deep:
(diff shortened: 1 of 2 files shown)


kmichel-aiven (changes requested):
Can you add a test for it so that we don't reintroduce the bug ?

rominf:
@kmichel-aiven What kind of test do you think is suitable? Mocking `format_key_for_backend` and verifying arguments that were passed? Sounds strange. I think systest against real swift object storage would be great, but we don't have systests, and it's not trivial.

kmichel-aiven:
> @kmichel-aiven What kind of test do you think is suitable? Mocking `format_key_for_backend` and verifying arguments that were passed? Sounds strange. I think systest against real swift object storage would be great, but we don't have systests, and it's not trivial to implement them.

Integration test with a real Swift, we can't do them yet (but soon).

I wouldn't mock `format_key_for_backend` because it's our own code and an implementation detail and whatever we pass it is no proof of what we end up passing to Swift.

I think mocking the Swift object `self.conn` and making sure that the prefix we actually pass to swift is as expected would be better.

kmichel-aiven (approved):

PR #72, opened 2022-09-22
Add some unit tests for object storages that were lacking them

heikju:

Add some tests for the object storage types that didn't have them yet. so that the recently added object storage notification feature gets better test coverage.

Resolves: #xxxxx

# Why this way


hackaugusto (approved):
LGTM

PR #75, opened 2022-09-30
pass object metadata also via the http notifier

sjamgade:
this helps notification consuming side not have query it if needed.

Resolves: #xxxxx

# Why this way

### DIFF
--- a/rohmu/notifier/http.py
+++ b/rohmu/notifier/http.py
@@ -130,7 +130,7 @@ class BackgroundHTTPNotifier(Notifier):
         self._stop_event.set()
         self._thread.join(_THREAD_JOIN_TIMEOUT)
 
-    def object_created(self, key: str, size: Optional[int]) -> None:
+    def object_created(self, key: str, size: Optional[int], metadata: Optional[dict]) -> None:
         self._queue.put(
             HTTPNotifyJob(
                 self._url,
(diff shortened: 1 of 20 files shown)


rohmu/notifier/http.py
```diff
@@ -130,7 +130,7 @@ def close(self) -> None:
         self._stop_event.set()
         self._thread.join(_THREAD_JOIN_TIMEOUT)
 
-    def object_created(self, key: str, size: Optional[int]) -> None:
+    def object_created(self, key: str, size: Optional[int], metadata: Optional[dict]) -> None:
```
fingon, line 133: Perhaps this signature could just have mandatory key and optional kwargs that default to None? With that, the callers would look bit cleaner.

rohmu/notifier/null.py
```diff
@@ -9,7 +9,7 @@ class NullNotifier(Notifier):
     Used by default if configuration is missing to avoid None checks
     """
 
-    def object_created(self, key: str, size: Optional[int]) -> None:
+    def object_created(self, key: str, size: Optional[int], metadata=None) -> None:
```
fingon, line 12: This doesn't quite match typing of the parent class.

rohmu/notifier/http.py
sjamgade: yes, the signature is making the calls smelly. Ideally, these would be dedicated types (`NotifyObjectCreated`, ....) (probably eventually when the API requirements stabilizes). 
kwargs for me is a classic way to forget passing additional arguments even if they are available.

fingon (approved):

PR #78, opened 2022-11-01
notfier: bump retry times and backup_off for http notifer

sjamgade:
this gives the receiving side more time to recover fromr a restart

Resolves: #xxxxx

# Why this way


hackaugusto (approved):
lgtm

PR #79, opened 2022-11-22
Provide consistent progress metric callbacks

nicois:
While some methods already support a progress metric being passed to them, some do not. This makes it difficult to transition upstream projects to rely on progress-based metrics and alerting.

This PR enhances `rohmufile` to inject an argument - the number of bytes transmitted - to its callback, while maintaining backwards-compatibility. It also adds progress callback support to `store_file_from_disk`, which previously did not support it, and allows more implementations to invoke the progress callbacks in more situations. (For example, the `s3` backend now provides progress during a multi-part file upload, which is very common)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,13 +23,13 @@ jobs:
     steps:
 
       - id: checkout-code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           # Do not persist the token during execution of this job.
           persist-credentials: false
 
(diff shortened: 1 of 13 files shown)


yaolu-aiven (approved):
LGTM

alexanderfomichev (approved):
Looks good to me

sjamgade:
is there a reason why boto3 was explicitly instead of botocore. 
if this was because of the stubs then could this be `youtype/botocore-stubs`  used ?

PR #80, opened 2022-11-24
notifier: sanitize metadata before trying to json dump

sjamgade:
this is needed because metdata can contain complex objects like datetime

An easier way to fix this would be to move sanitize_metadata out of baseclass to utils and directly call it from notifer, but in case of azure the hypen(-) in keys needs to be replaced with underscore (_). and notifer has no idea what storage driver was is use and if hyphen replacement should be done.

Resolves: #xxxxx

# Why this way

### DIFF
--- a/test/test_object_storage_azure.py
+++ b/test/test_object_storage_azure.py
@@ -44,16 +44,19 @@ def test_store_file_from_disk(azure_module: ModuleType, get_blob_client: MagicMo
         notifier=notifier,
     )
     test_data = b"test-data"
+    metadata = {"Content-Length": len(test_data), "some-object": object()}
     upload_blob = MagicMock()
     get_blob_client.return_value = MagicMock(upload_blob=upload_blob)
 
     with NamedTemporaryFile() as tmpfile:
         tmpfile.write(test_data)
         tmpfile.flush()
-        transfer.store_file_from_disk(key="test_key1", filepath=tmpfile.name)
+        transfer.store_file_from_disk(key="test_key1", filepath=tmpfile.name, metadata=metadata)
 
     upload_blob.assert_called_once()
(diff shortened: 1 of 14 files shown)


test/test_object_storage_azure.py
```diff
@@ -75,7 +79,9 @@ def upload_side_effect(*args, **kwargs):  # pylint: disable=unused-argument
     upload_blob = MagicMock(wraps=upload_side_effect)
     get_blob_client.return_value = MagicMock(upload_blob=upload_blob)
 
-    transfer.store_file_object(key="test_key2", fd=file_object)
+    transfer.store_file_object(key="test_key2", fd=file_object, metadata=metadata)
 
     upload_blob.assert_called_once()
-    notifier.object_created.assert_called_once_with(key="test_key2", size=len(test_data), metadata=None)
+    notifier.object_created.assert_called_once_with(
+        key="test_key2", size=len(test_data), metadata=transfer.sanitize_metadata(metadata, replace_hyphen_with="_")
```
heikju, line 86: In a test, where we define what the metadata input is, I think we should have the expected sanitized metadata as a dict literal here. That way, we test that `sanitize_metadata()` produces the output we expect it to. As this is now, it seems to me that we're just testing that the method produces the output the function produces, which is less useful, and doesn't show clearly to the reader of this test how the metadata is transformed.

And, since the point of this PR is to support datetimes, then I'd like to include one in the unsanitized metadata input for `store_file_object()`. And same goes for other storage backend tests too (since the hyphen replacement makes the expected output to not be the same for all of them).

heikju (changes requested):

test/test_object_storage_azure.py
sjamgade: added test for datetime handling of `santizie_metadata`

heikju (approved):

PR #82, opened 2022-11-30
tests: verify compatibility with python 3.11

nicois:
Now python 3.11 is stable, we should ensure there are no incompatibilities or regressions when using this version.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,7 +57,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ["3.10", "3.9", "3.8", "3.7"]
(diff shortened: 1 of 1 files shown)


aris-aiven:
Snappy package hasn't been built for python 3.11 (see relevant issue https://github.com/andrix/python-snappy/issues/124).

Therefore CI would require the following packages to build the `python-snappy` package:
- `libsnappy-dev` for Debian/Ubuntu and
- `snappy-devel` RPM for RedHat-based systems.

fingon (changes requested):
GHA doesn't seem happy with this at least.

aris-aiven:
Fixed by #121

PR #83, opened 2022-12-01
handle missing size key in result

sjamgade:
the size is key missing and the documentation mentions the key to be present but in practice that doesnot seem to be the case, sometimes.

Try and avoid throwing an exception in such cases

Resolves: #xxxxx

# Why this way

### DIFF
--- a/rohmu/object_storage/google.py
+++ b/rohmu/object_storage/google.py
@@ -226,7 +226,8 @@ class GoogleTransfer(BaseTransfer):
                 sourceObject=source_object,
             )
             result = self._retry_on_reset(request, request.execute)
-            self.notifier.object_copied(key=destination_key, size=int(result["size"]), metadata=metadata)
+            if result.get("size", None) is not None:
+                self.notifier.object_copied(key=destination_key, size=int(result["size"]), metadata=metadata)
 
     def get_metadata_for_key(self, key):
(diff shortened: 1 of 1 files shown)


rohmu/object_storage/google.py
```diff
@@ -226,7 +226,8 @@ def copy_file(self, *, source_key, destination_key, metadata=None, **_kwargs):
                 sourceObject=source_object,
             )
             result = self._retry_on_reset(request, request.execute)
-            self.notifier.object_copied(key=destination_key, size=int(result["size"]), metadata=metadata)
+            if result.get("size", None) is not None:
+                self.notifier.object_copied(key=destination_key, size=int(result["size"]), metadata=metadata)
```
rikonen, line 230: In the one problem that was seen size was available but under key `original-file-size` or something like that. Should this support getting from the alternative keys?

sjamgade: `original-file-size` is file size before compression (it is under metadata). the notifier needs the size google storage sees, which would be `compressed-file-size` under metadata. but I dont think generalizing on this would be a good idea. 

Also I dont think this happens all the time as I never saw it in my dev-evn. Also skipping notification once in a while is not bad, because consolidation will take care of it

fingon (approved):

PR #85, opened 2022-12-06
add missing test for google and fix getting size of uploaded data

sjamgade:

Resolves: #xxxxx

# Why this way


rohmu/object_storage/google.py
```diff
@@ -390,7 +390,7 @@ def store_file_from_memory(self, key, memstring, metadata=None, extra_props=None
         upload = MediaIoBaseUpload(data, mimetype or "application/octet-stream", chunksize=UPLOAD_CHUNK_SIZE, resumable=True)
         sanitized_metadata = self.sanitize_metadata(metadata)
         result = self._upload(upload, key, sanitized_metadata, extra_props, cache_control=cache_control)
-        self.notifier.object_created(key=key, size=int(result.get("size", len(data))), metadata=sanitized_metadata)
+        self.notifier.object_created(key=key, size=int(result.get("size", len(memstring))), metadata=sanitized_metadata)
```
rikonen, line 393: Not sure why we ever try reading the size field from the result since the original input size is guaranteed to be the correct value and is always present.

rikonen (approved):

PR #90, opened 2023-02-16
pydantic models for transfer objects

sjamgade:
these models are not yet directly used in the classes but act as way to validate/annotate the inputs expected by these transfer classes

Resolves: #xxxxx

# Why this way


akudiyar (approved):
LGTM. Discussed in DM

PR #91, opened 2023-02-25
Handle cases where s3 might return non-existing object

derrix060:
There are some cases where the cloud provider implementation of s3 is not that consistent and it will list a non-existing object, for example OVH

Making it a draft because probably that's not desired


rikonen (commented):
This would make file listing in many cases prohibitively slow. If something like this is wanted it should be controlled by new parameter for the function, allowing caller to decide whether this is OK or not depending on the case.

fingon:
I merged the competing proposal which handles it correctly only when with_metadata is used (no extra overhead there).

giacomo-alzetta-aiven:
Going to close this since alternative was merged.

PR #93, opened 2023-02-28
initial statsd implementation

sjamgade:
this just adds the statsd client but does not take it into use


sjamgade:
almost everything is copied from astacus

fingon (approved):

PR #94, opened 2023-03-09
common: fix ProxyInfo typing

raphlopez:
Configure ProxyInfo model to allow aliasing and make ProxyType inherit from str for string comparisons.

Fixes https://github.com/aiven/rohmu/pull/90

# Why this way
Pydantic model config `allow_population_by_field_name ` needs to be specified to allow aliasing: https://docs.pydantic.dev/usage/model_config/

`ProxyInfo.type` is used in several string comparisons throughout the codebase. Since `ProxyType` doesn't inherit from `str`, those comparisons result in unintended behaviour.

For example:
```
>>> from enum import Enum
>>> class MyEnum(str, Enum):
...     state1 = 'state1'
...     state2 = 'state2'
...
>>> a = MyEnum.state1
>>> a
<MyEnum.state1: 'state1'>
>>> a == "state1"
True

>>> class MyEnum(Enum):
...     state1 = 'state1'
...     state2 = 'state2'
...
>>> a = MyEnum.state1
>>> a
<MyEnum.state1: 'state1'>
>>> a == 'state1'
False

### CODE REVIEW

**Review by songnon — APPROVED** [review]

**Comment by sjamgade** [comment]
This LGTM. We might have more raised eyebrows with `pass` been changed to `password`. 
the `model.dict()` can be updated to `model.dict(by_alias=True)`

```python
>>> d = {"host":"q23", "port":1232, "type":ProxyType.socks5, "pass":"1232"}
>>> p = ProxyInfo(**d)
>>> p.dict()
{'host': 'q23', 'port': 1232, 'type': 'socks5', 'user': None, 'password': '1232'}
>>> p.dict(by_alias=True)
{'host': 'q23', 'port': 1232, 'type': 'socks5', 'user': None, 'pass': '1232'}
```

sjamgade (changes requested):
requesting changes for the above comment

sjamgade:
I think we are fine with the second commit
pydantic does this by itself anyways, its just mypy being mypy here
```
>>> SnapshotFile(relative_path="rando.txt", file_size=100, stored_file_size=100, mtime_ns=100, hexdigest="adf")
SnapshotFile(relative_path=PosixPath('rando.txt'), file_size=100, stored_file_size=100, mtime_ns=100, hexdigest='adf', content_b64=None, should_be_bundled=False, missing_ok=True)
```

sjamgade (approved):

PR #96, opened 2023-03-13
metrics for s3

sjamgade:
commit 2 and 3 are essentially the same. commit 3's approach feels less intrusive. 
Commit messages have more info

### DIFF
--- a/rohmu/object_storage/azure.py
+++ b/rohmu/object_storage/azure.py
@@ -6,6 +6,7 @@ Copyright (c) 2022 Aiven, Helsinki, Finland. https://aiven.io/
 See LICENSE for details
 """
 from ..common.models import ProxyInfo, StorageModel
+from ..common.statsd import StatsdConfig
 from ..notifier.interface import Notifier
 from .base import IncrementalProgressCallbackType, ProgressProportionCallbackType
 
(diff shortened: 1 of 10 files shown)


rominf (commented):
I like a decorator approach more. LGTM, however, I have no enough context to approve it.

rohmu/object_storage/s3.py
```diff
@@ -412,6 +424,7 @@ def multipart_upload_file_object(
             start_of_part_upload = time.monotonic()
             while True:
                 attempts -= 1
+                self.stats.operation(StorageOperation.store_file_from_memory)
```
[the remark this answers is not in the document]
sjamgade: added

rohmu/object_storage/azure.py
```diff
@@ -76,9 +77,10 @@ def __init__(
         azure_cloud=None,
         proxy_info=None,
         notifier: Optional[Notifier] = None,
+        statsd_info: Optional[StatsdConfig] = None,
```
akudiyar, line 80: q: This field name looks a bit controversial to me. Why is it "info" and not "config"?

rohmu/object_storage/base.py
```diff
@@ -41,14 +42,15 @@ class Config:
...
-    def __init__(self, prefix, notifier: Optional[Notifier] = None) -> None:
+    def __init__(self, prefix, notifier: Optional[Notifier] = None, statsd_info: Optional[StatsdConfig] = None) -> None:
         self.log = logging.getLogger(self.__class__.__name__)
         if not prefix:
             prefix = ""
         elif prefix[-1] != "/":
             prefix += "/"
         self.prefix = prefix
         self.notifier = notifier or NullNotifier()
+        self.stats = StatsClient(statsd_info)
```
akudiyar, line 53: if `None` is passed to `parse_obj()`, an error like `expected dict not NoneType (type=type_error)` will be thrown

akudiyar (changes requested):
Two small comments

rohmu/object_storage/azure.py
sjamgade: we have proxy_info, so went with same things

rohmu/object_storage/base.py
sjamgade: StatsClient checks for statsd_info being `None` before proceding

akudiyar (approved):

PR #98, opened 2023-03-16
Fix the pypi publisher

nicois:
Using the current version of github's checkout action, tags are not properly synced, even when `fetch-depth: 0` is present.

Explicitly fetch the tag references to ensure the annotated tag check succeeds. Without this change, the object type of the reference is simply `commit`, and not `tag`.


alexanderfomichev (approved):
LGTM

PR #99, opened 2023-03-17
Skip file when key is not found in s3.

RommelLayco:
There are cases where the list_objects_v2 api returns objects that are not actually present in s3. This cause an error to happen. We should skip these "files"

Skip file that fail the head_object api

Resolves: #xxxxx

# Why this way


derrix060:
Similar to #91 :)

fingon (approved):

PR #100, opened 2023-03-17
improve google stats reporting

sjamgade:
somestimes these files are really small (<1MB) sometimes just a few bytes. These get reported as 5MB over even more.

with these (stored_file_from_memory, store_file_from_string) calls it is trivial to get the actual size of data, this helps steer not too far away from the actual sizes.

this might lead to undercounting in case of retries but with such small sizes this should be fine

Resolves: #xxxxx

# Why this way

### DIFF
--- a/rohmu/object_storage/google.py
+++ b/rohmu/object_storage/google.py
@@ -8,7 +8,7 @@ See LICENSE for details
 # pylint: disable=import-error, no-name-in-module
 
 from ..common.models import ProxyInfo, StorageModel, StorageOperation
-from ..common.statsd import StatsdConfig
+from ..common.statsd import StatsClient, StatsdConfig
 from ..dates import parse_timestamp
 from ..errors import FileNotFoundFromStorageError, InvalidConfigurationError
 from ..notifier.interface import Notifier
@@ -24,14 +24,23 @@ from .base import (
 from contextlib import contextmanager
 from googleapiclient.discovery import build
 from googleapiclient.errors import HttpError
-from googleapiclient.http import build_http, MediaFileUpload, MediaIoBaseDownload, MediaIoBaseUpload, MediaUpload
+from googleapiclient.http import (
+    build_http,
+    MediaDownloadProgress,
+    MediaFileUpload,
+    MediaIoBaseDownload,
+    MediaIoBaseUpload,
+    MediaUpload,
+    MediaUploadProgress,
+)
 from http.client import IncompleteRead
 from io import BytesIO, FileIO
 from oauth2client import GOOGLE_TOKEN_URI
 from oauth2client.client import GoogleCredentials
-from typing import Optional
+from typing import Optional, Tuple, Union
 
 import codecs
+import dataclasses
 import errno
 
 # NOTE: this import is not needed per-se, but it's imported here first to point the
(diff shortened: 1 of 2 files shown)


sjamgade:
we might not be caring about numbers in case of backup. But these becomes revelant in case of replication as there the $source and $destination matters

ilpon (commented):
I don't know enough about rohmu & metrics to review.

sjamgade:
it should be okay, as long as nothing seems absolutely wrong it should be okay to merge. in worst the numbers/metrics would be wrong, but nothing relies on these metrics right away.

sjamgade:
Also under-reporting is troublesome but overreporting (not by huge error) is also fine.

rohmu/object_storage/google.py
```diff
@@ -259,26 +305,32 @@ def copy_file(self, *, source_key, destination_key, metadata=None, **_kwargs):
...
                 self.notifier.object_copied(key=destination_key, size=int(result["size"]), metadata=metadata)
 
     def get_metadata_for_key(self, key):
         path = self.format_key_for_backend(key)
         with self._object_client(not_found=path) as clob:
-            return self._metadata_for_key(clob, path)
+            return self._metadata_for_key(clob, path)[0]
 
-    def _metadata_for_key(self, clob, key):
+    def _metadata_for_key(self, clob, key) -> Tuple[dict, int]:
```
akudiyar, line 318: nit: built-in `tuple` also works as the type hint

rohmu/object_storage/google.py
```diff
@@ -117,6 +118,53 @@ def base64_to_hex(b64val):
...
+    size: Union[None, int, float] = None
+    progress_prev: int = 0
+
+    def report(self, stats: StatsClient):
+        # reports the default.
+        # for sized operation, reporting after the operation is fine as _retry_on_reset will
+        # have reported the something already
+        # and if the operation eventually succeeds then report_status() or this will take care
+        # of it.
+        SIZED_OPERATIONS = {
```
akudiyar, line 149: nit: this dict can be on the class level

rohmu/object_storage/google.py
```diff
@@ -259,26 +305,32 @@ def copy_file(self, *, source_key, destination_key, metadata=None, **_kwargs):
...
+            reporter.report(self.stats)
+            result = self._retry_on_reset(request, request.execute, retry_reporter=reporter)
             if result.get("size", None) is not None:
                 self.notifier.object_copied(key=destination_key, size=int(result["size"]), metadata=metadata)
 
     def get_metadata_for_key(self, key):
         path = self.format_key_for_backend(key)
         with self._object_client(not_found=path) as clob:
-            return self._metadata_for_key(clob, path)
+            return self._metadata_for_key(clob, path)[0]
```
akudiyar, line 316: nit: a class for the result type would be more reader-friendly

rohmu/object_storage/google.py
```diff
@@ -259,26 +305,32 @@ def copy_file(self, *, source_key, destination_key, metadata=None, **_kwargs):
                 sourceBucket=self.bucket_name,
                 sourceObject=source_object,
             )
-            result = self._retry_on_reset(request, request.execute, operation=StorageOperation.copy_file)
+            reporter.report(self.stats)
```
akudiyar, line 308: nit: I see quite some places where you call `Reporter()`, then `.report()` and then `_retry_on_reset()`. It could be combined together in one method

akudiyar: Also, if we know all the possible operations, the reporter instances for these operations could be built-in (and cached?)
Like: `StorageOperation.copy_file.report(self.stats)`

akudiyar (approved):
Overall LGTM, there are some nitpicks for increasing readability, but the're up to you

PR #101, opened 2023-04-17
object_storage: API + operation changes

fingon:
- Add get_contents_to_file/string to base class

  (No API change, just implementation simpliciy)

  Ultimately only get_contents_to_fileobj is the interesting one (if we don't implement proper parallel multipart download anyway), and the other two can just use that.

- Add store_file_from_memory, store_file_from_disk to base class

  NOTE: Backward incompatible API change - require keyword arguments also in store_file_from_disk (all Aiven code is fine with it, and we want this going forward)

  NOTE: Handling of 'metadata' varied by functions (in some it was keyword argument, in others not); made it non-keyword third argument in all 3 store variants. (This is backwards compatible, as restrictions were just loosened.)

- operation tracking changes/fixes

  - documented the operations somewhat

  - added multipart abort

  - simplified the list too; we don't really care about the type of get_file that is used (for example); also storing multipart chunks on S3 is just implementation detail and part of store_file API call

  - change operation so that it has both count and size metric (size one is populated only for metrics supplied with size)

  - use increase for both, the old gauge for size dropped data if there was more than one entry per metrics interval

This fixes some errors in statsd reporting, and cleans the API implementation quite a bit (while mostly being backwards compatible).

# Why this way

There was internal demand for statsd operation reporting fixes, and the API cleanup seemed like good idea at same time (most of providers are still not supported for statsd stuff, but it can be done easier after this).

### DIFF
--- a/rohmu/object_storage/base.py
+++ b/rohmu/object_storage/base.py
@@ -7,13 +7,16 @@ See LICENSE for details
 """
 from ..common.models import StorageModel
 from ..common.statsd import StatsClient, StatsdConfig
-from ..errors import StorageError
+from ..errors import FileNotFoundFromStorageError, StorageError
 from ..notifier.interface import Notifier
 from ..notifier.null import NullNotifier
 from collections import namedtuple
-from typing import Callable, Generic, Optional, Type, TypeVar
+from contextlib import suppress
+from io import BytesIO
+from typing import Callable, Generic, Optional, Type, TypeVar, Union
 
 import logging
+import os
 import platform
 
 KEY_TYPE_OBJECT = "object"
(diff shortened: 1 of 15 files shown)


ilpon (commented):
Looks good to me, but needs someone else to review.

kmichel-aiven (approved):
A few more parameter and return type hints on the modified functions would have been good, but there is less code that before, so still technically better than yesterday.

rohmu/object_storage/base.py
```diff
@@ -175,29 +206,71 @@ def sanitize_metadata(self, metadata, replace_hyphen_with="-"):
...
+                metadata = metadata.copy()
+                metadata["Content-Length"] = size
+            self.store_file_object(
+                key,
+                fd,
+                cache_control=cache_control,
+                metadata=metadata,
+                mimetype=mimetype,
+                multipart=multipart,
+                upload_progress_fn=self._incremental_to_proportional_progress(cb=progress_fn, size=size),
```
rikonen, line 262: This looks to be incorrect behavior for `S3Transfer` as that reports the total number of bytes sent every time instead of delta.

rikonen (changes requested):

rikonen (approved):
True, I didn't expand the lines and missed the `_proportional_to_incremental_progress` call there, just checked the `multipart_upload_file_object` implementation

Issue #13, opened 2022-04-27
Add type hints


carobme:
Add type hints to rohmu, as well as a mypy configuration.

PR #104, opened 2023-05-03
Enable stricter MyPy typing for Rohmu

giacomo-alzetta-aiven:

This PR enables various strict MyPy settings and adds/changes a lot of type hints to enable better typing support.

Some improvements can be added later on. For example, I/O typing is particularly messy in general. The current best practice would be to define a couple `Protocol`s and use that as arguments, why leaving the return types as concrete `io` classes. Unfortunately this is not simple because some stubs/library use the `typing.IO*` stuff which is not 100% compatible. Most Rohmu entities deal with `BufferedIOBase`, and so I took that as most general I/O thing, with a couple of `cast`s or `Union` in a few places where the `typing.IO` is inevitable.

Resolves: #13 

# Why this way

The API has not changed, except for the better types, but I expect that if you get a `TypeError` with the new version you probably have a bug. In a few places some plain attributes are now `@property`s, so they can't be set anymore.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,6 +1,7 @@
 [mypy]
 python_version = 3.9
 warn_redundant_casts = True
+disallow_any_generics = true
 
 
 [mypy-azure.*]
--- a/mypy.ini
+++ b/mypy.ini
@@ -2,6 +2,7 @@
 python_version = 3.9
 warn_redundant_casts = True
 disallow_any_generics = true
+disallow_subclassing_any = true
 
 
 [mypy-azure.*]
@@ -13,8 +14,6 @@ ignore_missing_imports = True
 [mypy-dataclasses.*]
 ignore_missing_imports = True
 
-[mypy-googleapiclient.*]
-ignore_missing_imports = True
 
 [mypy-oauth2client.*]
 ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -3,6 +3,7 @@ python_version = 3.9
 warn_redundant_casts = True
 disallow_any_generics = true
 disallow_subclassing_any = true
+disallow_untyped_decorators = true
 
 
 [mypy-azure.*]
--- a/mypy.ini
+++ b/mypy.ini
@@ -5,6 +5,7 @@ disallow_any_generics = true
 disallow_incomplete_defs = true
 disallow_subclassing_any = true
 disallow_untyped_decorators = true
+disallow_untyped_defs = true
 
 
 [mypy-azure.*]
@@ -28,4 +29,3 @@ ignore_missing_imports = True
 
 [mypy-systemd.*]
 ignore_missing_imports = True
-
--- a/mypy.ini
+++ b/mypy.ini
@@ -2,6 +2,7 @@
 python_version = 3.9
 warn_redundant_casts = True
 disallow_any_generics = true
+disallow_incomplete_defs = true
 disallow_subclassing_any = true
 disallow_untyped_decorators = true
 
--- a/rohmu/__init__.py
+++ b/rohmu/__init__.py
@@ -8,8 +8,8 @@ See LICENSE for details
 from .common.models import StorageModel
 from .errors import InvalidConfigurationError
 from .notifier.interface import Notifier
-from .object_storage.base import BaseTransfer
-from typing import Any, Dict, Mapping, Type
+from .object_storage.base import BaseTransfer, StorageModelT
+from typing import Any, cast, Dict, Mapping, Type, TypeVar
 
 IO_BLOCK_SIZE = 2**20  # 1 MiB
 STORAGE_TYPE = "storage_type"
@@ -17,37 +17,37 @@ NOTIFIER_TYPE = "notifier_type"
 Config = Mapping[str, Any]
 
 
-def get_class_for_transfer(obj_store: Config) -> Type[BaseTransfer]:
+def get_class_for_transfer(obj_store: Config) -> Type[BaseTransfer[StorageModel]]:
     storage_type = obj_store[STORAGE_TYPE]
     if storage_type == "azure":
         from .object_storage.azure import AzureTransfer
 
-        return AzureTransfer
+        return cast(Type[BaseTransfer[StorageModel]], AzureTransfer)
     elif storage_type == "google":
         from .object_storage.google import GoogleTransfer
 
-        return GoogleTransfer
+        return cast(Type[BaseTransfer[StorageModel]], GoogleTransfer)
     elif storage_type == "sftp":
         from .object_storage.sftp import SFTPTransfer
 
-        return SFTPTransfer
+        return cast(Type[BaseTransfer[StorageModel]], SFTPTransfer)
     elif storage_type == "local":
         from .object_storage.local import LocalTransfer
 
-        return LocalTransfer
+        return cast(Type[BaseTransfer[StorageModel]], LocalTransfer)
     elif storage_type == "s3":
         from .object_storage.s3 import S3Transfer
 
-        return S3Transfer
+        return cast(Type[BaseTransfer[StorageModel]], S3Transfer)
     elif storage_type == "swift":
         from .object_storage.swift import SwiftTransfer
 
-        return SwiftTransfer
+        return cast(Type[BaseTransfer[StorageModel]], SwiftTransfer)
 
     raise InvalidConfigurationError("unsupported storage type {0!r}".format(storage_type))
 
 
-def get_class_for_notifier(notifier_config: dict) -> Type[Notifier]:
+def get_class_for_notifier(notifier_config: dict[str, Any]) -> Type[Notifier]:
     notifier_type = notifier_config[NOTIFIER_TYPE]
     if notifier_type == "http":
         from .notifier.http import BackgroundHTTPNotifier
@@ -56,14 +56,14 @@ def get_class_for_notifier(notifier_config: dict) -> Type[Notifier]:
     raise InvalidConfigurationError("unsupported storage type {0!r}".format(notifier_type))
 
 
-def get_notifier(notifier_config: dict) -> Notifier:
+def get_notifier(notifier_config: dict[str, Any]) -> Notifier:
     notificer_class = get_class_for_notifier(notifier_config)
     notifier_config = notifier_config.copy()
     notifier_config.pop(NOTIFIER_TYPE)
     return notificer_class(**notifier_config)
 
 
-def get_transfer_model(storage_config: Config) -> Type[StorageModel]:
+def get_transfer_model(storage_config: Config) -> StorageModel:
     storage_class = get_class_for_transfer(storage_config)
     storage_config = dict(storage_config)
     storage_config.pop(STORAGE_TYPE)
@@ -76,7 +76,7 @@ def get_transfer_model(storage_config: Config) -> Type[StorageModel]:
     return model
 
 
-def get_transfer(storage_config: Config) -> BaseTransfer:
+def get_transfer(storage_config: Config) -> BaseTransfer[StorageModel]:
     storage_class = get_class_for_transfer(storage_config)
     model = get_transfer_model(storage_config)
     return storage_class.from_model(model)
--- a/rohmu/__init__.py
+++ b/rohmu/__init__.py
@@ -8,13 +8,13 @@ See LICENSE for details
 from .common.models import StorageModel
 from .errors import InvalidConfigurationError
 from .notifier.interface import Notifier
-from .object_storage.base import BaseTransfer, StorageModelT
-from typing import Any, cast, Dict, Mapping, Type, TypeVar
+from .object_storage.base import BaseTransfer
+from typing import Any, cast, Dict, Type
 
 IO_BLOCK_SIZE = 2**20  # 1 MiB
 STORAGE_TYPE = "storage_type"
 NOTIFIER_TYPE = "notifier_type"
-Config = Mapping[str, Any]
+Config = Dict[str, Any]
 
 
 def get_class_for_transfer(obj_store: Config) -> Type[BaseTransfer[StorageModel]]:
@@ -47,7 +47,7 @@ def get_class_for_transfer(obj_store: Config) -> Type[BaseTransfer[StorageModel]
     raise InvalidConfigurationError("unsupported storage type {0!r}".format(storage_type))
 
 
-def get_class_for_notifier(notifier_config: dict[str, Any]) -> Type[Notifier]:
+def get_class_for_notifier(notifier_config: Config) -> Type[Notifier]:
     notifier_type = notifier_config[NOTIFIER_TYPE]
     if notifier_type == "http":
         from .notifier.http import BackgroundHTTPNotifier
@@ -56,7 +56,7 @@ def get_class_for_notifier(notifier_config: dict[str, Any]) -> Type[Notifier]:
     raise InvalidConfigurationError("unsupported storage type {0!r}".format(notifier_type))
 
 
-def get_notifier(notifier_config: dict[str, Any]) -> Notifier:
+def get_notifier(notifier_config: Config) -> Notifier:
     notificer_class = get_class_for_notifier(notifier_config)
     notifier_config = notifier_config.copy()
     notifier_config.pop(NOTIFIER_TYPE)
--- a/rohmu/atomic_opener.py
+++ b/rohmu/atomic_opener.py
@@ -1,34 +1,18 @@
 from contextlib import contextmanager
 from pathlib import Path
-from typing import BinaryIO, Callable, cast, ContextManager, Generator, Optional, overload, TextIO, TYPE_CHECKING, Union
+from typing import BinaryIO, Callable, cast, ContextManager, Generator, Optional, overload, TextIO, Union
+from typing_extensions import TypeAlias
 
-# workaround < Python3.8 missing Literal.
-# if we just use the try/except block, mypy will complain
-# about the fact that Write* variables are being redefined.
-# so we need to pour in an additional branch.
-if TYPE_CHECKING:
+try:
     from typing import Literal
-
-    Write = Literal["w"]
-    WriteBinary = Literal["wb"]
-    WriteTest = Literal["somethingrandomw"]
-else:
-    try:
-        from typing import Literal
-
-        Write = Literal["w"]
-        WriteBinary = Literal["wb"]
-        WriteTest = Literal["somethingrandomw"]
-    except ImportError:
-        from typing import Any
-
-        Write = Any
-        WriteBinary = Any
-        WriteTest = Any
-
+except ImportError:
+    from typing_extensions import Literal  # type: ignore [assignment]
 
 import errno
 import os
+
+Write: TypeAlias = Literal["w"]
+WriteBinary: TypeAlias = Literal["wb"]
 
 
 def _fd_close_quietly(fd: int) -> None:
@@ -61,19 +45,6 @@ def atomic_opener(
     _after_link_hook: Callable[[], None] = lambda: None,
 ) -> ContextManager[TextIO]:
     ...
-
-
-if TYPE_CHECKING:
-    # necessary for testing or tests will fail type checking
-    @overload
-    def atomic_opener(
-        final_path: Path,
-        mode: WriteTest,
-        encoding: Optional[str] = None,
-        _fd_spy: Callable[[int], None] = lambda unused: None,
-        _after_link_hook: Callable[[], None] = lambda: None,
-    ) -> ContextManager[TextIO]:
-        ...
 
 
 def atomic_opener(
--- a/rohmu/common/statsd.py
+++ b/rohmu/common/statsd.py
@@ -14,9 +14,12 @@ This is combination of:
 - pydantic configuration + explicit typing
 
 """
+
+from __future__ import annotations
+
 from contextlib import asynccontextmanager, contextmanager
 from enum import Enum
-from typing import Any, AsyncIterator, Dict, Iterator, Optional, Union
+from typing import AsyncIterator, Dict, Iterator, Optional, Union
 
 import pydantic
 import socket
@@ -91,7 +94,7 @@ class StatsClient:
         self._send(metric, b"ms", value, tags)
 
     def unexpected_exception(self, ex: BaseException, where: str, *, tags: Optional[Tags] = None) -> None:
-        all_tags: dict[str, Any] = {
+        all_tags: Tags = {
             "exception": ex.__class__.__name__,
             "where": where,
         }
--- a/rohmu/common/statsd.py
+++ b/rohmu/common/statsd.py
@@ -16,7 +16,7 @@ This is combination of:
 """
 from contextlib import asynccontextmanager, contextmanager
 from enum import Enum
-from typing import Dict, Optional, Union
+from typing import Any, AsyncIterator, Dict, Iterator, Optional, Union
 
 import pydantic
 import socket
@@ -64,12 +64,12 @@ class StatsClient:
         self._operation_map = config.operation_map
 
     @asynccontextmanager
-    async def async_timing_manager(self, metric: str, tags: Optional[Tags] = None):
+    async def async_timing_manager(self, metric: str, tags: Optional[Tags] = None) -> AsyncIterator[None]:
         with self.timing_manager(metric, tags=tags):
             yield
 
     @contextmanager
-    def timing_manager(self, metric: str, tags: Optional[Tags] = None):
+    def timing_manager(self, metric: str, tags: Optional[Tags] = None) -> Iterator[None]:
         start_time = time.monotonic()
         tags = (tags or {}).copy()
         try:
@@ -90,15 +90,15 @@ class StatsClient:
     def timing(self, metric: str, value: Union[int, float], *, tags: Optional[Tags] = None) -> None:
         self._send(metric, b"ms", value, tags)
 
-    def unexpected_exception(self, ex, where, *, tags: Optional[Tags] = None) -> None:
-        all_tags = {
+    def unexpected_exception(self, ex: BaseException, where: str, *, tags: Optional[Tags] = None) -> None:
+        all_tags: dict[str, Any] = {
             "exception": ex.__class__.__name__,
             "where": where,
         }
         all_tags.update(tags or {})
         self.increase("exception", tags=all_tags)
 
-    def operation(self, operation, *, count: int = 1, size: Union[int, None] = None):
+    def operation(self, operation: str, *, count: int = 1, size: Union[int, None] = None) -> None:
         tags: Tags = {"operation": self._operation_map.get(str(operation), str(operation))}
         self.increase("rohmu_operation_count", tags=tags, inc_value=count)
         if size is not None:
--- a/rohmu/common/statsd.py
+++ b/rohmu/common/statsd.py
@@ -27,7 +27,7 @@ class MessageFormat(str, Enum):
     datadog = "datadog"
     telegraf = "telegraf"
 
-    def __str__(self):
+    def __str__(self) -> str:
         return str(self.value)
 
 
--- a/rohmu/compressor.py
+++ b/rohmu/compressor.py
@@ -7,11 +7,10 @@ See LICENSE for details
 from .errors import InvalidConfigurationError
 from .filewrap import Sink, Stream
 from .snappyfile import SnappyFile
-from .typing import Compressor, Decompressor, HasRead
+from .typing import BinaryData, Compressor, Decompressor, FileLike, HasRead, HasWrite
 from .zstdfile import open as zstd_open
-from typing import cast, IO, Union
+from typing import cast, IO
 
-import io
 import lzma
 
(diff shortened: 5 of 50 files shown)


rohmu/object_storage/google.py
```diff
@@ -426,17 +452,17 @@ def get_file_size(self, key):
...
-        cache_control,
+        upload: MediaUpload,
+        key: str,
+        metadata: Metadata,
+        extra_props: Optional[Dict[str, Any]],
+        cache_control: Optional[str],
         reporter: Reporter,
         upload_progress_fn: IncrementalProgressCallbackType = None,
-    ):
+    ) -> Any:
```
fingon, line 462: No need for this return value - I think google implementations store_file_object returning something is a bug, not a feature (others do not return anything and in general API should be what is in base.py).

rohmu/object_storage/google.py
```diff
@@ -505,16 +531,16 @@ def store_file_from_disk(
...
-        mimetype=None,
-        multipart: Union[bool, None] = None,
+        cache_control: Optional[str] = None,
+        mimetype: Optional[str] = None,
+        multipart: Optional[bool] = None,  # pylint: disable=unused-argument
         upload_progress_fn: IncrementalProgressCallbackType = None,
-        extra_props=None,  # pylint: disable=arguments-differ
-    ):  # pylint: disable=unused-argument
+        extra_props: Optional[Dict[str, Any]] = None,  # pylint: disable=arguments-differ
+    ) -> Any:
```
fingon, line 543: about all other backends return None, so should this.

rohmu/typing.py
```diff
@@ -0,0 +1,58 @@
...
+        ...
+
+
+class HasRead(Protocol):
+    def read(self, n: Optional[int] = -1) -> bytes:
+        ...
+
+
+class Compressor(Protocol):
+    def compress(self, data: bytes) -> Any:
```
fingon, line 49:
```suggestion
    def compress(self, data: bytes) -> bytes:
```

fingon (changes requested):
Overall good work. Some comments.

In general what 'most' of object storages do should be API in base.py, and then all object storages should strictly implement that particular API (in terms of signature). If there's divergence, it is most likely bug and not feature.

rohmu/encryptor.py
```diff
@@ -21,7 +23,7 @@
 import os
 import struct
 
-if cryptography.__version__ < "1.6":  # type: ignore
+if cryptography.__version__ < "1.6":
```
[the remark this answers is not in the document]
giacomo-alzetta-aiven: Yeah. rohmu.spec currently has `python3-cryptography >= 0.8` which is from 2015, version 1.6 is from the end of 2016 so I don't think many people are using current rohmu with that old cryptography versions.

rohmu/common/statsd.py
```diff
@@ -90,15 +93,15 @@ def increase(self, metric: str, *, inc_value: int = 1, tags: Optional[Tags] = No
     def timing(self, metric: str, value: Union[int, float], *, tags: Optional[Tags] = None) -> None:
         self._send(metric, b"ms", value, tags)
 
-    def unexpected_exception(self, ex, where, *, tags: Optional[Tags] = None) -> None:
-        all_tags = {
+    def unexpected_exception(self, ex: BaseException, where: str, *, tags: Optional[Tags] = None) -> None:
+        all_tags: dict[str, Any] = {
```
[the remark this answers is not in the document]
giacomo-alzetta-aiven: This one does not work:

```
python3 -m mypy --show-error-codes rohmu/ test/
rohmu/common/statsd.py:101: error: Argument 1 to "update" of "MutableMapping" has incompatible type "Dict[str, Union[int, str, None]]"; expected "SupportsKeysAndGetItem[str, str]"  [arg-type]
rohmu/common/statsd.py:102: error: Argument "tags" to "increase" of "StatsClient" has incompatible type "Dict[str, str]"; expected "Optional[Dict[str, Union[int, str, None]]]"  [arg-type]
rohmu/common/statsd.py:102: note: "Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
rohmu/common/statsd.py:102: note: Consider using "Mapping" instead, which is covariant in the value type
```

Sumit defined the `Tags` as `Dict[str, Union[int, str, None]]`. I have no idea where these tags come from and what values can they have but here they can be `str`, `int` or `None`. I can change the `Any` with `Union[str, int, None]` though.

rohmu/atomic_opener.py
```diff
@@ -66,7 +65,7 @@ def atomic_opener(
 if TYPE_CHECKING:
```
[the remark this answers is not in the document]
giacomo-alzetta-aiven: Yes, without this running on Python3.7 fails with:
```
ImportError: cannot import name 'Literal' from 'typing' (/usr/lib64/python3.7/typing.py)
```
(e.g. executing the unittests).

But I think this can be simplified a lot. I believe the current implementation was done to avoid using any type comments... but IMHO if we can use 4 lines instead of 20 with a type comment and when moving to python3.8 simply remove the workaround I think it's better.

rohmu/object_storage/base.py
```diff
@@ -164,59 +175,63 @@ def get_contents_to_file(self, key, filepath_to_store_to, *, progress_callback:
...
         # This method isn't currently used by PGHoard itself, it is merely provided
         # for applications that use PGHoard's object storage abstraction layer.
         raise NotImplementedError
 
-    def get_metadata_for_key(self, key):
+    def get_metadata_for_key(self, key: str) -> Metadata:
         raise NotImplementedError
 
-    def list_path(self, key, *, with_metadata=True, deep=False):
+    def list_path(self, key: str, *, with_metadata: bool = True, deep: bool = False) -> list[Any]:
```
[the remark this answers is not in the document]
giacomo-alzetta-aiven: @fingon  The problem here is that `iter_key` returns a sequence of `IterKeyItem` which is defined as:

```
IterKeyItem = namedtuple("IterKeyItem", ["type", "value"])
```

so the `value` attribute is `Any` . Maybe we should redefine this to be generic in some way so we can use the type variable to provide a better result type? Or can you tell me if the `value` is actually always fixed?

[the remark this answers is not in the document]
giacomo-alzetta-aiven: Okay, so we could use this definition:

```
from typing import TypeVar, Generic, NamedTuple
Value = TypeVar("Value")


class IterKeyItem(NamedTuple, Generic[Value]):
    type: str
    value: Value
```

But then all uses of `IterKeyItem` will need a type argument . I guess the `Value` type depends on the provider?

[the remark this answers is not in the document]
fingon: no, value is str and type is enum (either TYPE_PREFIX or TYPE_OBJECT)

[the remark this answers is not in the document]
giacomo-alzetta-aiven: @fingon  Are you sure? 

In base.py inside `BaseTransfer.delete_tree` this is what I find:
```
names = [item["name"] for item in self.list_path(key, with_metadata=False, deep=True)]
```

and `list_path` is just a wrapper for `list_iter`:
```
    def list_path(self, key: str, *, with_metadata: bool = True, deep: bool = False) -> list[Any]:
        return list(self.list_iter(key, with_metadata=with_metadata, deep=deep))
```

Which yields the `value`s from `iter_key`:

```
    def list_iter(self, key: str, *, with_metadata: bool = True, deep: bool = False) -> Iterator[Any]:
        for item in self.iter_key(key, with_metadata=with_metadata, deep=deep):
            if item.type == KEY_TYPE_OBJECT:
                yield item.value
```

So `value` can't always be `str`, otherwise the `item["name"]` in delete tree would fail.

Or am I missing something?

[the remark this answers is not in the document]
fingon: Ah, right, the value's type depends on the type - PREFIX is str and OBJECT is indeed (platform dependant) dict. Even for those, we should really define fixed TypedDict but perhaps not worth it yet.

rohmu/compressor.py
```diff
@@ -23,10 +25,10 @@
     zstd = None  # type: ignore
 
 
-def CompressionFile(dst_fp, algorithm, level=0, threads=0):
+def CompressionFile(dst_fp: io.BufferedIOBase, algorithm: str, level: int = 0, threads: int = 0) -> io.BufferedIOBase:
```
fingon, line 28:
```suggestion
def CompressionFile(dst_fp: HasRead, algorithm: str, level: int = 0, threads: int = 0) -> io.BufferedIOBase:
```

rohmu/compressor.py
```diff
@@ -54,17 +57,17 @@ def __init__(self, src_fp, algorithm, level=0):
...
+    def _process_chunk(self, data: bytes) -> bytes:
         return self._compressor.compress(data)
 
-    def _finalize(self):
+    def _finalize(self) -> Any:
         return self._compressor.flush()
 
 
-def DecompressionFile(src_fp, algorithm):
+def DecompressionFile(src_fp: io.BufferedIOBase, algorithm: str) -> io.BufferedIOBase:
```
fingon, line 67:
```suggestion
def DecompressionFile(src_fp: HasRead, algorithm: str) -> io.BufferedIOBase:
```

rohmu/encryptor.py
```diff
@@ -59,93 +67,111 @@ def update(self, data):
...
-        return len(data)
+        self.offset += len(data_as_bytes)
+        return len(data_as_bytes)
 
 
 class EncryptorStream(Stream):
     """Non-seekable stream of data that adds encryption on top of given source stream"""
 
-    def __init__(self, src_fp, rsa_public_key_pem):
+    def __init__(self, src_fp: io.BufferedIOBase, rsa_public_key_pem: Union[str, bytes]) -> None:
```
fingon, line 128:
```suggestion
    def __init__(self, src_fp: HasRead, rsa_public_key_pem: Union[str, bytes]) -> None:
```

rohmu/encryptor.py
```diff
@@ -208,23 +244,23 @@ def _reset(self):
         self.state = "OPEN"
 
     @classmethod
-    def _file_size(cls, file):
+    def _file_size(cls, file: io.BufferedIOBase) -> int:
```
fingon, line 247: nit: this is really weird classmethod and not your problem but it could be really something outside class too.

rohmu/compressor.py
giacomo-alzetta-aiven: Yeah, changes this actually requires more changes downstream to FileWrap & all the various subclasses. `read` is not enough for them.

I believe we use at least: `read`, `write`, `flush`, `close` and `fileno`, possibly other methods in that hierarchy so I guess we need an other protocol for that other than `HasRead`.

I'm going to try and see what we end up with.

fingon (commented):
Looks good to me now (+- unhappiness with formatting so not approving it yet). I'd like to get second opinion for this though.

giacomo-alzetta-aiven:
I rebased the PR to avoid conflicts in rohmufile.py.

mypy.ini
```diff
@@ -1,6 +1,11 @@
 [mypy]
 python_version = 3.9
```
rominf, line 2: I believe we should use `python_version = 3.7` here, since it's the lowest supported version for the project. To make typing more powerful and annotations easier, you can add an additional dependency: https://pypi.org/project/typing-extensions/3.10.0.2/ (this exact version, to be bumped to https://pypi.org/project/typing-extensions/4.4.0/ in near future).

giacomo-alzetta-aiven: It probably makes sense to change this python 3.9 to either 3.7 or 3.10. Not sure what @fingon  thinks about this. 

About the suggestion of checking python 3.11 discussed in the daily today, apparently the snappy package does not yet support 3.11 and from this issue it seems likely that they might never support it since the current maintainer is suggesting moving to cramjam... So moving Rohmu to python3.11 might have that issue.

If we need to change this I would prefer moving to 3.10 since 3.7 is EOL in ~1.5 months, and we should probably make sure that the unittest we at least import all modules (to catch runtime errors due to annotation definitions).
I will add the typing extensions module in the requirements explicitly.

giacomo-alzetta-aiven: Okay, the typing_extensions package is already provided because it is required by a number of other packages:
```
$ pip show typing_extensions
Name: typing_extensions
Version: 4.5.0
Summary: Backported and Experimental Type Hints for Python 3.7+
Home-page: 
Author: 
Author-email: "Guido van Rossum, Jukka Lehtosalo, Łukasz Langa, Michael Lee" <levkivskyi@gmail.com>
License: 
Location: /home/giacomo.alzetta/Projects/rohmu/venv/lib/python3.10/site-packages
Requires: 
Required-by: azure-core, azure-storage-blob, google-api-python-client-stubs, mypy, mypy-boto3-sqs, pydantic
```
I believe it makes sense to include it explicitly. I would directly use version 4.5.0 which still supports python3.7.

fingon: 3.7 is really soon EOL (c.f. https://endoflife.date/python) so I wouldn't go for that at least, 3.8 might make sense though.

rohmu/encryptor.py
```diff
@@ -34,21 +30,33 @@ class EncryptorError(Exception):
...
-        self.authenticator = None
+        self.rsa_public_key = cast(
+            RSAPublicKey, serialization.load_pem_public_key(rsa_public_key_pem, backend=default_backend())
+        )
+        self._cipher: Optional[CipherContext] = None
+        self._authenticator: Optional[HMAC] = None
+
+    @property
+    def cipher(self) -> CipherContext:
+        assert self._cipher is not None
```
bugant, line 44: nit: Wouldn't be better to raise a proper exception here instead of using `assert`? (like you did in https://github.com/aiven/rohmu/pull/104/files#diff-5c190a4f634c25382cbcfbc9997d46e5668951c55b395c4082beb4c7e3077312R92)

rohmu/encryptor.py
```diff
@@ -59,93 +67,111 @@ def update(self, data):
...
+        return self._cipher
+
+    def expected_header_bytes(self) -> int:
         if self._header_size is not None:
             return 0
         return self._cipher_key_len or 8
 
-    def header_size(self):
+    def header_size(self) -> int:
+        assert self._header_size is not None, "Header size was not initialized"
```
bugant, line 168: nit: Also here, I'd use a proper execption within an `if`

rohmu/object_storage/azure.py
```diff
@@ -156,36 +162,41 @@ def copy_file(self, *, source_key, destination_key, metadata=None, **kwargs):
...
-                item_name = item.value
+                item_name = cast(str, item.value)
             if item_name.rstrip("/").rsplit("/", 1)[-1] == expected_name:
                 break
         else:
             item = None
         if not item or item.type != KEY_TYPE_OBJECT:
             raise FileNotFoundFromStorageError(path)  # not found or prefix
-        return item.value["metadata"]
+        return cast(Dict[str, Any], item.value)["metadata"]
```
bugant, line 187: nit: Is it possible to cast to `Metadata` instead of `Dict[str, Any]`?

bugant (approved):
It looks good to me.  I left a couple of minor (non-blocking) comments.

mypy.ini
giacomo-alzetta-aiven: IMHO, python3.7 is getting EOLed so I wouldn't want to move back there and next month having to change this again. For now I'm leaving the weird setup. In a next step we should probably move to python3.10.

For Python3.11 there is an issue with the snappy dependency which will need to be resolved...

rohmu/object_storage/azure.py
giacomo-alzetta-aiven: Yes, but here the cast is on the "wrapper" object from which we extract the metadata.

Technically `Metadata` is just `Dict[str, Any]` but since I'm using it to refer to whatever is found under the `metadata` key I want to avoid using it outside that context. In the future the `Metadata` might become some `TypedDict` and suddenly it would not work correctly as used here.

rohmu/encryptor.py
giacomo-alzetta-aiven: I decided to go with a custom `UninitializedException`.

rohmu/object_storage/azure.py
```diff
@@ -67,13 +71,13 @@ class AzureTransfer(BaseTransfer[Config]):
...
-        prefix=None,
-        azure_cloud=None,
-        proxy_info=None,
+        bucket_name: str,
+        account_name: str,
+        account_key: Optional[str] = None,
+        sas_token: Optional[str] = None,
+        prefix: Optional[str] = None,
+        azure_cloud: Optional[str] = None,
+        proxy_info: Optional[dict[str, str]] = None,
```
sjamgade, line 80: this is not entirely correct,
proxy_info has `port` which should be of type `int`

use `ProxyInfo` instead ?

sjamgade (commented):
noticed some loose assumptions. It would be nice to have them corrected in this or a follow up PR.

Havent looked at stuff under delta/* but rest LGTM

rohmu/notifier/http.py
```diff
@@ -130,7 +132,7 @@ def close(self) -> None:
         self._stop_event.set()
         self._thread.join(_THREAD_JOIN_TIMEOUT)
 
-    def object_created(self, key: str, size: Optional[int], metadata: Optional[dict]) -> None:
+    def object_created(self, key: str, size: Optional[int], metadata: Optional[dict[str, Any]]) -> None:
```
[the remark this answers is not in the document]
giacomo-alzetta-aiven: Yes, you are right. I just checked and it seems all places that call this pass through `sanitize_metadata`. I will fix this soon.

rohmu/__init__.py
```diff
@@ -9,45 +9,45 @@
...
 
 IO_BLOCK_SIZE = 2**20  # 1 MiB
 STORAGE_TYPE = "storage_type"
 NOTIFIER_TYPE = "notifier_type"
-Config = Mapping[str, Any]
+Config = Dict[str, Any]
 
 
-def get_class_for_transfer(obj_store: Config) -> Type[BaseTransfer]:
+def get_class_for_transfer(obj_store: Config) -> Type[BaseTransfer[StorageModel]]:
```
akudiyar, line 20: Have you considered using "tagged unions" for typing these transfer objects? 
See https://docs.pydantic.dev/latest/usage/types/#discriminated-unions-aka-tagged-unions
It seems better than casting to a "base" type

rohmu/atomic_opener.py
```diff
@@ -53,7 +37,7 @@ def atomic_opener(
 
 
 @overload
-def atomic_opener(
+def atomic_opener(  # type: ignore [misc]
```
akudiyar, line 40: q: Why do we have to ignore that, what was the error?

rohmu/delta/common.py
```diff
@@ -76,7 +81,7 @@ class Config:
         # possibly the tests themselves are broken
         validate_assignment = True
 
-    def jsondict(self, **kw):
+    def jsondict(self, **kw: Any) -> dict[str, Any]:
```
akudiyar, line 84: `Dict`?

rohmu/inotify.py
```diff
@@ -52,20 +58,20 @@ def parse_inotify_buffer(event_buffer):
 
 
 class InotifyWatcher(Thread):
-    def __init__(self, compression_queue):
+    def __init__(self, compression_queue: Queue[dict[str, str]]) -> None:
```
akudiyar, line 61: `Dict`? Or maybe `Mapping` is better since it's an inbound partameter

rohmu/inotify.py
```diff
@@ -42,7 +48,7 @@ class InotifyEvent(ctypes.Structure):
 IN_NONBLOCK = 0x00004000
 
 
-def parse_inotify_buffer(event_buffer):
+def parse_inotify_buffer(event_buffer: bytes) -> Iterator[tuple[int, int, int, bytes]]:
```
akudiyar, line 51: Some docstring with output explanation or better a named tuple could be useful here

rohmu/notifier/http.py
```diff
@@ -85,7 +87,7 @@ def background_http_request(
 
 
 def initialize_background_thread(
-    queue: Queue,
+    queue: "Queue[HTTPNotifyJob]",
```
akudiyar, line 90: Will `from __future__ import annotations` help with not using a string hint here?

rohmu/object_storage/azure.py
```diff
@@ -67,13 +71,13 @@ class AzureTransfer(BaseTransfer[Config]):
...
     def __init__(
         self,
-        bucket_name,
-        account_name,
-        account_key=None,
-        sas_token=None,
-        prefix=None,
-        azure_cloud=None,
-        proxy_info=None,
+        bucket_name: str,
```
akudiyar, line 74: It seems that we need named parameters in this constructor, because most of them are optional

rohmu/notifier/null.py
```diff
@@ -9,7 +12,7 @@ class NullNotifier(Notifier):
     Used by default if configuration is missing to avoid None checks
     """
 
-    def object_created(self, key: str, size: Optional[int], metadata: Optional[dict]) -> None:
+    def object_created(self, key: str, size: Optional[int], metadata: Optional[dict[str, Any]]) -> None:
```
akudiyar: `Dict`?

rohmu/notifier/logger.py
```diff
@@ -1,14 +1,17 @@
...
-from typing import Optional
+from typing import Any, Optional
 
 
 class LoggerNotifier(Notifier):
     def __init__(self, log: Logger) -> None:
         self._log = log
 
-    def object_created(self, key: str, size: Optional[int], metadata: Optional[dict]) -> None:
+    def object_created(self, key: str, size: Optional[int], metadata: Optional[dict[str, Any]]) -> None:
```
akudiyar: `Dict`?

rohmu/object_storage/azure.py
```diff
@@ -156,36 +162,41 @@ def copy_file(self, *, source_key, destination_key, metadata=None, **kwargs):
...
         expected_name = path.rsplit("/", 1)[-1]
+        item: Optional[IterKeyItem]
         for item in items:
+            assert item is not None  # for mypy
             # We expect single result but Azure listing is prefix match so we need to explicitly
             # look up the matching result
+            item_name: str
             if item.type == KEY_TYPE_OBJECT:
-                item_name = item.value["name"]
+                item_name = cast(Dict[str, str], item.value)["name"]
```
akudiyar, line 178: It seems better to check the type here instead of casting, isn't it?

akudiyar (changes requested):
Not a full review because the PR is refreshing often, but still some questions

akudiyar (dismissed):
Not a full review because the PR is refreshing often, but still some questions

rohmu/atomic_opener.py
giacomo-alzetta-aiven: @akudiyar I believe this is something I had added to make mypy work with 3.7 setting as suggested by Roman, but since in the end I decided to just keep 3.9 and let a future PR decide which version we should target I guess this could be removed together with a few other comments.

rohmu/delta/common.py
giacomo-alzetta-aiven: There is a mix of `Dict` vs `dict` here but it's mostly due to python 3.7 support where `dict` does not support arguments.

Since I added the `from __future__ import annotations` flag in a few flag there is no issue with using `dict` with annotations.

There are a couple of places where we *must* use `Dict` (e.g. defining a type alias).

I guess I should replace all uses of `Dict` in annotations with `dict` and only leave `Dict` for type aliases until we drop python3.7 support.

rohmu/notifier/http.py
giacomo-alzetta-aiven: Probably. I certainly added this before adding the future import. So I will review all these cases after lunch.

rohmu/object_storage/azure.py
giacomo-alzetta-aiven: IMHO since you cannot do `isinstance(..., Dict[str, str])` checking that `item.value` is a `dict` plus that the value associated to key `name` is a `str` is going to be more complicated.

rohmu/__init__.py
giacomo-alzetta-aiven: That seems like work for a different PR IMHO... Here I just fixed the type to use stricter mypy settings which was the goal of this PR. What you suggest seems to require some changes in the class hierarchy to make it work.

rohmu/object_storage/azure.py
fingon: Public API -> please no signature change unless really needed.

rohmu/compressor.py
```diff
@@ -23,10 +24,10 @@
     zstd = None  # type: ignore
 
 
-def CompressionFile(dst_fp, algorithm, level=0, threads=0):
+def CompressionFile(dst_fp: FileLike, algorithm: str, level: int = 0, threads: int = 0) -> FileLike:
```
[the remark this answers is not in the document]
giacomo-alzetta-aiven: The level values depends on the compression algorithm. Most have a parameter from 0 to 9, ZSTD has a range 1-22. But, IMHO, I don't think that using `Literal[int]` here would give us anything. It would only make harder to write code that uses this function without much benefits. 

The `Literal[str]` in the `atomic_opener` makes much more sense because the `@overload` can be used to specify `TextIO` vs `BinaryIO` in return type, but whether the compression level is `1`, `100` or 1 million doesn't really change much and seems only a configuration parameter, so I'd rather leave this as plain `int`.

rohmu/delta/snapshot.py
```diff
@@ -45,13 +49,13 @@ class Snapshotter:
     def __init__(
         self,
         *,
-        src,
-        dst,
-        globs,
+        src: StrOrPathLike,
+        dst: StrOrPathLike,
+        globs: list[str],
```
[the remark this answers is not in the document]
giacomo-alzetta-aiven: They are available from python3.9. Using deferred annotations you can import the code even in python3.7, even though it has limitations (e.g. cannot use this in Pydantic model fields).  We are running mypy for python3.9 atm.

IMHO after merging this we can easily add a PR to just replace all uses of List/Tuple etc with list/tuple where possible. I don't really want to delay this PR further for this kind of change.

rohmu/inotify.py
giacomo-alzetta-aiven: That would be the only use of Mapping in rohmu. For this PR I'd prefer using the built-in, if later we want to use Mapping and other more generic types we can go hunt all the places where we could change those and do so.

rohmu/object_storage/azure.py
```diff
@@ -156,36 +162,46 @@ def copy_file(self, *, source_key, destination_key, metadata=None, **kwargs):
...
                 item_name = item.value["name"]
             else:
+                assert isinstance(item.value, str)
                 item_name = item.value
             if item_name.rstrip("/").rsplit("/", 1)[-1] == expected_name:
                 break
         else:
             item = None
         if not item or item.type != KEY_TYPE_OBJECT:
             raise FileNotFoundFromStorageError(path)  # not found or prefix
```
fingon, line 188: nit perhaps, but this is more readable for me at least:
```suggestion
        for item in items:
            # We expect single result but Azure listing is prefix match so we need to explicitly
            # look up the matching result
            item_name: str
            if item.type == KEY_TYPE_OBJECT:
                assert isinstance(item.value, dict)
                item_name = item.value["name"]
            else:
                assert isinstance(item.value, str)
                item_name = item.value
            if item_name.rstrip("/").rsplit("/", 1)[-1] == expected_name:
                break
        else:
            raise FileNotFoundFromStorageError(path)
        if item.type != KEY_TYPE_OBJECT:
            raise FileNotFoundFromStorageError(path)
```

fingon (changes requested):
Sorry, still perhaps one or two worth changing (and rest nit-ish).

rohmu/encryptor.py
```diff
@@ -59,93 +70,114 @@ def update(self, data):
...
-        self.encryptor = Encryptor(self.key)
+        self._encryptor: Optional[Encryptor] = Encryptor(self.key)
         self.offset = 0
         self.state = "OPEN"
 
-    def flush(self):
+    @property
+    def encryptor(self) -> Encryptor:
+        if self._encryptor is None:
+            raise ValueError("encryptor was not initialized!")
```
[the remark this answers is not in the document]
giacomo-alzetta-aiven: Fixed

rohmu/encryptor.py
```diff
@@ -163,41 +195,59 @@ def process_header(self, data):
...
+    def crypted_size(self) -> int:
+        if self._crypted_size is None:
+            raise UninitializedError("crypted_size not initialized")
+        return self._crypted_size
+
+    @property
+    def plaintext_size(self) -> int:
+        if self._plaintext_size is None:
+            raise UninitializedError("plaintext_size not initialized")
+        return self._plaintext_size
```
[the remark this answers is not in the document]
giacomo-alzetta-aiven: You are right. In the other places the cipher/authenticator etc were already public. Fixed this.

fingon (approved):

PR #106, opened 2023-05-16
object_storage: batching for S3 delete_objects

jsou:

Fixes when S3 would receive a request to delete a large number of objects.

# Why this way

Without batching, any requests to delete_objects containing more than 1000 objects would fail (the [current limit ](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/delete_objects.html)).


giacomo-alzetta-aiven (approved):
This LGTM.

giacomo-alzetta-aiven:
I will let @fingon  merge this since I'm not sure about the process for merging rohmu PRs.

fingon (approved):

fingon:
Merging here is entirely normal; it is just internal use packaging that is extra step.

PR #107, opened 2023-05-17
s3: Do nothing when delete_keys is supplied with empty collection

fingon:


fingon:
Actually this is no longer needed - the batched change fixes this too (it doesn't do empty batches).

PR #108, opened 2023-05-17
Miscellaneous fixes

fingon:

- add support for hashes in local storage (useful when testing against it, as most storages report it)

- add primitive delete tests

### DIFF
--- a/rohmu/object_storage/local.py
+++ b/rohmu/object_storage/local.py
@@ -23,12 +23,15 @@ from typing import Any, BinaryIO, Iterator, Optional, TextIO, Union
 
 import contextlib
 import datetime
+import hashlib
 import json
 import os
 import shutil
 import tempfile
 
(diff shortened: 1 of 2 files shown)


rohmu/object_storage/local.py
```diff
@@ -207,18 +213,23 @@ def store_file_object(
         target_path = self.format_key_for_backend(key.strip("/"))
         os.makedirs(os.path.dirname(target_path), exist_ok=True)
         bytes_written = 0
+        m = hashlib.sha256()
```
giacomo-alzetta-aiven, line 216: nit: this is a bit weird, but from what I see it's already done for, e.g., azure etags and other stuff so ok for me.

I see that SFTP also lacks the hash field.

giacomo-alzetta-aiven (approved):
LGTM

PR #109, opened 2023-05-17
Implement byte_range download

giacomo-alzetta-aiven:

It allows providing a byte_range for the method `get_contents_to_fileobj` of transfer classes.

# Why this way

Some APIs provide built-in support for this change, others do not.

For the Google implementation:
This PR uses a patched version of a googleapiclient class to add support to the byte_ranges.

I've looked a bit in the alternative of moving to google's cloud libraries, but to me it sounds like a lot of work and I don't see a simple way to use those just for the use-case of adding byte-range support so it would mean a rewrite of the whole google implementation from what I see.

BTW: I started from mst-dev branch but I rebased it on `main` fixing the conflicts, that's why there are some fingon's commits.

### DIFF
--- a/rohmu/object_storage/azure.py
+++ b/rohmu/object_storage/azure.py
@@ -17,7 +17,7 @@ from .base import IncrementalProgressCallbackType, ProgressProportionCallbackTyp
 # pylint: disable=import-error, no-name-in-module
 from azure.core.exceptions import HttpResponseError, ResourceExistsError
 from azure.storage.blob import BlobServiceClient, ContentSettings
-from typing import Any, BinaryIO, Iterator, Optional, Tuple, Union,
+from typing import Any, BinaryIO, Iterator, Optional, Tuple, Union
 
 import azure.common
 import logging
@@ -254,7 +254,13 @@ class AzureTransfer(BaseTransfer[Config]):
 
         return int(content_range.split(" ", 1)[1].split("/", 1)[1])
 
-    def _stream_blob(self, key: str, fileobj: BinaryIO, byte_range: Optional[tuple[int, int]], progress_callback: ProgressProportionCallbackType) -> None:
+    def _stream_blob(
+        self,
+        key: str,
+        fileobj: BinaryIO,
+        byte_range: Optional[tuple[int, int]],
+        progress_callback: ProgressProportionCallbackType,
+    ) -> None:
         """Streams contents of given key to given fileobj. Data is read sequentially in chunks
         without any seeks. This requires duplicating some functionality of the Azure SDK, which only
         allows reading entire blob into memory at once or returning data from random offsets"""
@@ -300,6 +306,7 @@ class AzureTransfer(BaseTransfer[Config]):
         progress_callback: ProgressProportionCallbackType = None,
     ) -> Metadata:
         path = self.format_key_for_backend(key, remove_slash_prefix=True)
+        self._validate_byte_range(byte_range)
 
         self.log.debug("Starting to fetch the contents of: %r", path)
         try:
--- a/rohmu/object_storage/azure.py
+++ b/rohmu/object_storage/azure.py
@@ -17,7 +17,7 @@ from .base import IncrementalProgressCallbackType, ProgressProportionCallbackTyp
 # pylint: disable=import-error, no-name-in-module
 from azure.core.exceptions import HttpResponseError, ResourceExistsError
 from azure.storage.blob import BlobServiceClient, ContentSettings
-from typing import Any, BinaryIO, Iterator, Optional, Union
+from typing import Any, BinaryIO, Iterator, Optional, Tuple, Union,
 
 import azure.common
 import logging
@@ -229,6 +229,7 @@ class AzureTransfer(BaseTransfer[Config]):
                         "metadata": metadata,
                         "name": self.format_key_from_backend(item.name),
                         "size": item.size,
+                        "md5": item.etag.strip('"'),
                     },
                 )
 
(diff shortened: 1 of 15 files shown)


rohmu/object_storage/azure.py
```diff
@@ -253,21 +254,27 @@ def _parse_length_from_content_range(cls, content_range: str) -> int:
...
-        start_range = 0
+        start_range = byte_range[0] if byte_range else 0
         chunk_size = self.conn._config.max_chunk_get_size  # type: ignore [attr-defined] # pylint: disable=protected-access
         end_range = chunk_size - 1
         blob = self.conn.get_blob_client(self.container_name, key)
         while True:
             try:
                 # pylint: disable=protected-access
-                download_stream = blob.download_blob(offset=start_range, length=chunk_size)
+                if byte_range:
```
sjamgade, line 275: this should make sure byte_range[1] > byte_range[0]

rohmu/object_storage/google.py
```diff
@@ -703,3 +701,92 @@ def _read_bytes(self, length: int, *, initial_data: Optional[bytes] = None) -> b
             return read_results[0]
         else:
             return b"".join(read_results)
+
+
+class MediaIoBaseDownload:
```
sjamgade, line 706: prefer using the built-in one when byte_range is not provided.

Also iiuc, to get byte_range download just setting `_progress` would get use what we need ?
it is an internal attribute of the class but that is more preferrable than having a copy of that class here

https://googleapis.github.io/google-api-python-client/docs/epy/googleapiclient.http-pysrc.html#MediaIoBaseDownload

sjamgade: plus a test, which creates the subclass and mocks out the retry_request method to checks if the correct headers are making it over, would be needed.

sjamgade (changes requested):
got some subjective comments

rohmu/object_storage/google.py
giacomo-alzetta-aiven: @sjamgade I had already seen that it would be possible to hack around the implementation overriding the fields. 

I don't think `_progress` is enough though, there is also `_total_size` which needs to be worked around to report the correct progress and done status: https://github.com/googleapis/google-api-python-client/blob/main/googleapiclient/http.py#L765

We would have to have `_total_size` as some property/descriptor so that we can intercept here https://github.com/googleapis/google-api-python-client/blob/main/googleapiclient/http.py#L761 and replace it with the byte range size when needed... and that's really ugly.

Or we have to recompute all those fields from scratch in every call... and at that point you have rewritten half the class.

From my PV it's better to have our own implementation that works how we want, so we are sure that it won't ever change behaviour if we do a bugfix upgrade of the library.

If you want I can create an other branch with this more hackish solution and see how that works out, but my guess is that we don't save much code and make it more complicated and less stable due to possible internal changes in the library.

rohmu/object_storage/azure.py
giacomo-alzetta-aiven: Probably yes. I only worked on the google part, the other implementations are from Markus... maybe I should give a look to ensure we have proper validation & tests for this for the other implementations too.

rohmu/object_storage/google.py
giacomo-alzetta-aiven: @sjamgade  I have create a separate draft PR for what would look like messing with the internals in a subclass: https://github.com/aiven/rohmu/pull/110  . It is a bit shorter but to me this PR looks clearer.

If you think we should go with the other approach let me know and I will polish that one (I just hacked the implementation in the shortest time possible).

sjamgade: true, this is cleaner given the details

giacomo-alzetta-aiven: I'm going to clean up a few things and make sure we use our implementation only when the byte range is provided and then I will re-ask for review then.

rohmu/object_storage/azure.py
sjamgade: I would put a helper and have the byte_range validated across all the drivers. At quite a few places the assumption needs to hold or things go wonky.

rohmu/object_storage/google.py
```diff
@@ -703,3 +723,98 @@ def _read_bytes(self, length: int, *, initial_data: Optional[bytes] = None) -> b
...
+
+        total_size: Optional[int] = None
+        if resp.status in (200, 206):
+            if "content-location" in resp and resp["content-location"] != self._uri:
+                self._uri = resp["content-location"]
+            self._progress += len(content)
+            self._fd.write(content)
+
+            if "content-range" in resp:
+                total_size = get_total_size_from_content_range(resp["content-range"])
```
sjamgade, line 801: is there is a reason content-range would not be in response ?

rohmu/object_storage/google.py
```diff
@@ -703,3 +723,98 @@ def _read_bytes(self, length: int, *, initial_data: Optional[bytes] = None) -> b
...
+        Returns:
+          (status, done): The value of done will be True when the media has been fully
+             downloaded or the total size of the media is unknown.
+
+        Raises:
+          googleapiclient.errors.HttpError if the response was not a 2xx (or a 416 is received and the file is empty)
+          httplib2.HttpLib2Error if a transport error has occurred.
+        """
+        headers = self._headers.copy()
+        chunk_start = self._progress + self._start_range
```
sjamgade, line 786:
```suggestion
        chunk_start = max(self._progress, self._start_range)
```
chunk_start should start from `self._start_range` only for the first chunk, then its always `self._position`

sjamgade (dismissed):
almost there

rohmu/object_storage/google.py
giacomo-alzetta-aiven: This handles also 200 OKs.

From the specs it is valid for a server to return a 200 OK with the whole object if it makes sense to do so/the server/a proxy in between does not support Range (e.g. I ask range `0-100` for some content and the content is 10 bytes, the server can either return a 206 with `Content-Range 0-10/10` or a `200 OK` with `Content-Length: 10` and no `Content-Range` required from what I understand).

This handling of 200/206 and these headers is taken from the library so it should be correct or we need to bugfix the official implementation too.

rohmu/object_storage/google.py
sjamgade: > If you think about multiple calls to this functions, the _start_range will always be the same, while _progress increases. What you suggest would end up duplicating some bytes, because it will first return _start_range until _progress is bigger and then return that instead.

I might be mis-undestanding, but `self._position` gets advanced after every chunk is downloaded (`self._progress += len(content)`) . then adding `self._start_range` (to that, on next call to `next_chunk`) would be wrong ?

giacomo-alzetta-aiven: I have renamed the variables to make the meaning clearer. There is no `_position` only `_num_bytes_downloaded`, which at the beginning is 0. When you want to download a chunk of the response you add `_num_bytes_downloaded` to the `_start_position` of the range to get the current position you are at where the next chunk starts, so `chunk_start = self._num_bytes_downloaded + self._start_position`.

Does this make sense to you?

sjamgade: this was making it weird.
```
# pseudo code
chunk_start = self._progress + self._start_position
...
download(start_at=chunk_start)
..
self._progress += len(content))
```
but the latest change has fixed it.

rohmu/object_storage/google.py
```diff
@@ -402,25 +404,43 @@ def delete_key(self, key: str) -> None:
...
         last_log_output = 0.0
+        self._validate_byte_range(byte_range)
         with self._object_client(not_found=path) as clob:
             metadata, obj_size = self._metadata_for_key(clob, path)
-            reporter = Reporter(StorageOperation.get_file, size=min(obj_size, DOWNLOAD_CHUNK_SIZE))
+            if byte_range is None:
+                size_to_download = obj_size
+            else:
+                size_to_download = min(
+                    obj_size - byte_range[0], byte_range[1] - byte_range[0] + 1 if byte_range else obj_size
```
fingon, line 425: `if byte_range` here looks spurious?

fingon (commented):
LGTM, one nit (and as I was involved I'm not going to be the one officially reviewing and approving it anyway).

rohmu/object_storage/google.py
giacomo-alzetta-aiven: Yes, I guess originally I was handling everything in a single expression and I forgot to remove that condition.

PR #111, opened 2023-05-25
object_storage: remove boto3 dependency

lionbee:
Remove the boto3 dependency, there are no functional changes, 

# Why this way
The boto3 dependency is not needed, since we also use the botocore dependency. Additionally the boto3 dependency is more likely to result in dependency mismatches.

### DIFF
--- a/rohmu/object_storage/s3.py
+++ b/rohmu/object_storage/s3.py
@@ -24,15 +24,47 @@ from .base import (
 )
 from botocore.response import StreamingBody
 from rohmu.util import batched
-from typing import Any, BinaryIO, Collection, Iterator, Optional, Tuple, Union
+from typing import Any, BinaryIO, cast, Collection, Iterator, Optional, Tuple, TYPE_CHECKING, Union
 
-import boto3
 import botocore.client
 import botocore.config
 import botocore.exceptions
 import botocore.session
 import math
 import time
+
+if TYPE_CHECKING:
+    from mypy_boto3_s3 import S3Client
+
+
+# botocore typing stubs are incomplete. We either have to write all the stubs we need
+# or work around the problem a bit by importing the typing library only for type checking purposes.
+def create_s3_client(
+    *,
+    session: botocore.session.Session,
+    config: botocore.config.Config,
+    aws_access_key_id: Optional[str],
+    aws_secret_access_key: Optional[str],
+    aws_session_token: Optional[str],
+    region_name: str,
+    verify: Optional[bool | str] = None,
+    endpoint_url: Optional[str] = None,
+) -> S3Client:
+    s3_client = session.create_client(
+        "s3",
+        config=config,
+        aws_access_key_id=aws_access_key_id,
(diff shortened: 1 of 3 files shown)


rohmu/object_storage/s3.py
```diff
@@ -24,16 +24,50 @@
...
 import time
 
+# botocore typing stubs are incomplete. We either have to write all the stubs we need
+# or work around the problem a bit by importing the typing library only for type checking purposes.
+if TYPE_CHECKING:
+    from mypy_boto3_s3 import S3Client
+
+    s3_client_type = S3Client
+else:
+    s3_client_type = Any
```
lionbee, line 43: I feel a bit dirty for doing this, but it does seem to have the expected effect

fingon: I don't think it is needed, as with future annotation import, the annotations are not evaluated anyway but instead are strings. Due to that, in non TYPE_CHECKING they should not be used.

rohmu/object_storage/s3.py
```diff
@@ -24,16 +24,50 @@
...
+    session: botocore.session.Session,
+    config: botocore.config.Config,
+    aws_access_key_id: Optional[str],
+    aws_secret_access_key: Optional[str],
+    aws_session_token: Optional[str],
+    region_name: str,
+    verify: Optional[bool | str] = None,
+    endpoint_url: Optional[str] = None,
+) -> s3_client_type:
+    return cast(
```
fingon, line 57: cast is not relevant for non-type-checking case anyway; this whole block could happen under if TYPE_CHECKING? And return value could be always S3Client (even if it is not defined).

fingon (changes requested):

rohmu/object_storage/s3.py
lionbee: Because this needs `mypy_boto3_s3` there will have to be a conditional import and cast. Alternatively we will have to install `mypy_boto3_s3` during runtime as well and that seems wrong for this scenario. I did tidy up the logic a bit more so that no cast is required at runtime and no special type variables are defined

rohmu/object_storage/s3.py
```diff
@@ -35,6 +34,37 @@
...
+    *,
+    session: botocore.session.Session,
+    config: botocore.config.Config,
+    aws_access_key_id: Optional[str],
+    aws_secret_access_key: Optional[str],
+    aws_session_token: Optional[str],
+    region_name: str,
+    verify: Optional[bool | str] = None,
+    endpoint_url: Optional[str] = None,
+) -> Any:
```
fingon, line 49:
```suggestion
) -> S3Client:
```

and import it somewhere on toplevel if TYPE_CHECKING. With Any here, mypy won't help.

lionbee: ah right... I got what you wanted :rocket:

nicois (approved):
This is good. Maybe further tweaks are possible but let's lock in this forward progress

PR #112, opened 2023-05-30
object_storage: Ensure size is always an int

RommelLayco:
In the case of OVH they are returning this value as str. Ensure this value is always an int

### DIFF
--- a/rohmu/object_storage/base.py
+++ b/rohmu/object_storage/base.py
@@ -108,7 +108,7 @@ class BaseTransfer(Generic[StorageModelT]):
             # We could actually sniff from fd if it is seekable; left TODO for now.
             return default
 
-        return size > chunk_size
+        return int(size) > chunk_size
 
     @classmethod
     def from_model(cls, model: StorageModelT) -> BaseTransfer[StorageModelT]:


rohmu/object_storage/base.py
```diff
@@ -108,7 +108,7 @@ def _should_multipart(
             # We could actually sniff from fd if it is seekable; left TODO for now.
             return default
```
lionbee, line 109: `default` is of type `bool`

RommelLayco: yeah this returns a boolean

lionbee (approved):
LGTM, I'm an idiot :D

rohmu/object_storage/base.py
```diff
@@ -108,7 +108,7 @@ def _should_multipart(
...
 
-        return size > chunk_size
+        return int(size) > chunk_size
```
fingon, line 111: This is the wrong place to do it and against type hint. It should be fixed in source (wherever that is).

fingon (changes requested):

bridgetb (approved):

PR #113, opened 2023-06-12
implement transfer pooling

giacomo-alzetta-aiven:

Implements a transfer-pool API that allows multiple user threads to safely share rohmu transfer objects.

A new `rohmu.transfer_pool` module was added which defines `TransferPool` class. The class provides 2 methods to get a transfer object:

* The `with_transfer(storage_config)` method can be used as context manager
* The `get_transfer(storage_config)` method can be used to obtain a transfer object. This is a proxy that will act exactly as the normal transfer and includes a `return_to_pool` method which MUST be called by the user once they are done. All interactions after this call are prevented.

The module also provides a default pool instance which can be used via the `with_transfer` and `get_transfer` global functions.

# Why this way

Various object storage backends are not thread-safe, so users would need to manually handle thread-safety.
This API allows for a simple way to efficiently handle transfer objects in a thread-safe way.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,7 +57,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ["3.10", "3.9", "3.8", "3.7"]
+        python-version: ["3.10", "3.9", "3.8"]
 
     steps:
       - id: checkout-code
--- a/rohmu/transfer_pool.py
+++ b/rohmu/transfer_pool.py
@@ -1,0 +1,219 @@
+"""
+rohmu - transfer_pool
+
+Copyright (c) 2023 Aiven, Helsinki, Finland. https://aiven.io/
+See LICENSE for details
+"""
+
+from __future__ import annotations
+
+from . import get_transfer as rohmu_get_transfer
+from .errors import InvalidTransferError
+from .object_storage.base import BaseTransfer, StorageModel
+from contextlib import contextmanager
+from typing import Any, Callable, Generator, Optional
+
+import heapq
+import json
+import logging
+import threading
+import time
+
+LOG = logging.getLogger(__name__)
+
+TRANSFER_MAX_AGE = 60 * 60 * 3
+TRANSFER_CACHE_MAX_AGE = TRANSFER_MAX_AGE * 2
+
+
+class TransferCacheItem:
+    create_time: float
+    cache_key: str
+    transfer: BaseTransfer[StorageModel]
+    max_age: float
+
+    def __init__(self, pool_key: str, transfer: BaseTransfer[StorageModel], max_age: float = TRANSFER_MAX_AGE) -> None:
+        self.create_time = time.monotonic()
+        self.cache_key = pool_key
+        self.transfer = transfer
+        self.max_age = max_age
+
+    def age(self) -> float:
+        return time.monotonic() - self.create_time
+
+    def is_expired(self) -> bool:
+        return self.age() > self.max_age
+
+    def __gt__(self, other: TransferCacheItem) -> bool:
+        """Ordering by creation time, needed for storing these in a heap."""
+        return self.create_time > other.create_time
+
+
+class _TransferCache:
+    transfers_heap: list[TransferCacheItem]
+    last_used: float
+    max_age: float
+
+    def __init__(self, max_age: float = TRANSFER_CACHE_MAX_AGE) -> None:
+        self.transfers_heap = []
+        self.last_used = time.monotonic()
+        self.max_age = max_age
+
+    def prune_expired(self) -> None:
+        while self.transfers_heap and self.transfers_heap[0].is_expired():
+            heapq.heappop(self.transfers_heap)
+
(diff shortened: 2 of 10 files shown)


.github/workflows/build.yml
```diff
@@ -57,7 +57,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ["3.10", "3.9", "3.8", "3.7"]
+        python-version: ["3.10", "3.9", "3.8"]
```
giacomo-alzetta-aiven, line 60: Maybe this should be an other PR, but I wanted to use `:=` in my code... let me know if I should split this one in 2.

rohmu/transfer_pool.py
```diff
@@ -0,0 +1,220 @@
+"""
+rohmu - transfer_pool
```
giacomo-alzetta-aiven, line 2: Implementation follows the robjecter one. Differences:
* Using a class instead of just globals
* Provide context-manager interface
* Use a proxy for the "get" interface to try and help finding bugs related to users forgetting the "return" call to the pool
* Optimizing the pool if backend is thread-safe
* minor refactorings

rohmu/transfer_pool.py
```diff
@@ -0,0 +1,220 @@
...
+    def get(self) -> Optional[TransferCacheItem]:
+        self.last_used = time.monotonic()
+        self.prune_expired()
+        if self.transfers_heap:
+            try:
+                # there will always be only one transfer instance at most in this cache type
+                return self.transfers_heap[0]
+            except IndexError:
+                pass
+        return None
```
fingon, line 104:
```suggestion
    """

    def get(self) -> Optional[TransferCacheItem]:
        self.last_used = time.monotonic()
        if self.transfers_heap:
            try:
                # there will always be only one transfer instance at most in this cache type
                return self.transfers_heap[0]
            except IndexError:
                pass
        return None
```

Expiration of transfers in single transfer case isn't really meaningful. For multi transfer cache, it matters, but here we'd be recreating identical object and that is just waste of resources (presumably per-request stuff isn't long lived anyway, and even if it was, deleting it just before recreating it probably won't help much).

rohmu/transfer_pool.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def __init__(self, item: TransferCacheItem, pool_reclaim: Callable[[TransferCacheItem], None]) -> None:
+        self._item = item
+        self._pool_reclaim = pool_reclaim
+        self._done = False
+
+    def __del__(self) -> None:
+        if not self._done:
+            LOG.warning("Transfer not marked as done was garbage collected. Returning it to the pool now.")
+            self.return_to_pool()
```
fingon, line 133: Probably logging only here would be good enough too - not like we really care about keeping the transfer.

rohmu/transfer_pool.py
```diff
@@ -0,0 +1,220 @@
...
+        """Yield a transfer object according to the provided storage_config.
+
+        The transfer object may be created or cached. While the context manager is running the transfer object
+        won't be available to other threads.
+
+        Users of this contextmanager should NOT use the transfer object outside the `with` block
+        """
+        transfer = self._get(storage_config)
+        yield transfer.transfer
+        self._put(transfer)
```
fingon, line 180:
```suggestion
        transfer = self._get(storage_config)
        try:
            yield transfer.transfer
        finally:
            self._put(transfer)
```

rohmu/transfer_pool.py
```diff
@@ -0,0 +1,220 @@
...
+        self._caches = {key: pool for key, pool in self._caches.items() if not pool.is_expired()}
+
+    def _get(self, storage_config: dict[str, Any]) -> TransferCacheItem:
+        cache_key = _cache_key(storage_config)
+        with self._mutex:
+            if (cache := self._caches.get(cache_key)) and (cached_transfer := cache.get()):
+                return cached_transfer
+            transfer = rohmu_get_transfer(storage_config)
+            pool_cls = _TransferCacheForThreadSafeTransfer if transfer.is_thread_safe else _TransferCache
+            self._caches[cache_key] = pool_cls(max_age=self._max_pool_age)
```
fingon, line 202: This could be really created only on put I guess (it would also eliminate the 'no transfer pool' branch there that is there now).

fingon (dismissed):
Mostly LGTM. Probably the context manager thing is only really urgent one, few other minor comments too.

fingon:
The single transfer pool really can create up to N if they're parallel, which is bit unexpected. I'd probably just put it to the cache immediately once created (which sort of invalidates my comment about putting the transfer cache creation in put, I guess).

rohmu/transfer_pool.py
giacomo-alzetta-aiven: Removed expiration in that pool and made sure we always have a single instance.

rohmu/transfer_pool.py
giacomo-alzetta-aiven: Fixed, also added a test checking that it works in case of exceptions

fingon (approved):

PR #114, opened 2023-06-14
Bump to version 1.1.0

giacomo-alzetta-aiven:

Bumping to version 1.1.0

Somehow main is at 1.0.9 but PyPI has version 1.0.10 available... I guess in the last release the tag was pushed and the branch deleted instead of being merged into main?

# Why this way

I need a new version so that we can update robjecter dependency on rohmu to use new transfer pool.

After merging this to main I will push the tag to trigger the new release.

### DIFF
--- a/rohmu/version.py
+++ b/rohmu/version.py
@@ -1,1 +1,1 @@
-VERSION = "1.0.9"
+VERSION = "1.1.0"


sjamgade:
after looking at the changes, we dropped 3.7 (officially) which would warrant a major bump

fingon (changes requested):
Sumit has a point - I think this should be really 1.1 as it is backwards incompatible in a sense that it drops Python 3.7 support.

giacomo-alzetta-aiven:
> Sumit has a point - I think this should be really 1.1 as it is backwards incompatible in a sense that it drops Python 3.7 support.

I'm not sure. In the end we do mark the versions supported so pip used on python 3.7 won't ever try to install the incompatible version... that's only an issue for people manually installing it... but I don't know how relevant is semantic versioning.

In any case I guess I can change this to `1.1.0` or `2.0.0` as you like.

fingon (approved):

sjamgade (approved):

PR #115, opened 2023-06-28
local: fix broken backward compatibility for files without md5 metadata

giacomo-alzetta-aiven:

The local driver now considers the hash of the file as optional metadata.

# Why this way

If a user used a previous version of rohmu to save some files, those will not have the hash in the metadata stored. If they used the new rohmu version to list the files it will barf at the missing md5.

We could compute it on the fly but for now simply ignore it if it is missing.

### DIFF
--- a/rohmu/typing.py
+++ b/rohmu/typing.py
@@ -1,19 +1,8 @@
 from __future__ import annotations
 
 from types import TracebackType
-from typing import Any, Dict, Optional, Type, TYPE_CHECKING, Union
-
-try:
-    from typing import Protocol
-except ImportError:
-    from typing_extensions import Protocol  # type: ignore [assignment]
-
-try:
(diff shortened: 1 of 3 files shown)


rohmu/typing.py
```diff
@@ -25,15 +14,7 @@
...
-BinaryData = Union[
-    bytes,
-    bytearray,
-    memoryview,
-    "array[Any]",
-    mmap.mmap,
-    "ctypes._CData",
-    PickleBuffer,
-]
+BinaryData = Buffer
```
giacomo-alzetta-aiven, line 17: Weird since we had mypy set to python3.9 since well before, but for some reason it now complains about overriding the type in FileWrap if we don't do this...

sjamgade (approved):

PR #120, opened 2023-07-05
Revert "typing: fix liskov principle errors"

giacomo-alzetta-aiven:

This reverts commit 5c058b759d64fb08919d1f2034e4ae647d4d0e16.

It broke code using older typing_extensions versions.

Instead we use a couple of `type: ignore` comments to stop mypy from complaining.

# Why this way

### DIFF
--- a/rohmu/encryptor.py
+++ b/rohmu/encryptor.py
@@ -113,7 +113,7 @@ class EncryptorFile(FileWrap):
         self._check_not_closed()
         return True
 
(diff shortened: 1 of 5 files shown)


aris-aiven:
Woudl it be possible to rebase on main and re-run the actions? (just to make sure, I trust your changes)

aris-aiven (approved):
LGTM

PR #121, opened 2023-07-05
Add python3.11 pipeline to the build

giacomo-alzetta-aiven:

Build rohmu also for python3.11

# Why this way

We need to install `libsnappy-dev` because otherwise `python-snappy` won't install on python3.11


aris-aiven (approved):
Great, thanks for taking the time to do this

PR #122, opened 2023-07-05
lint: use Python 3.11 for the workflow

aris-aiven:

Uses Python 3.11 for the static analysis workflow to ensure main builds & checks against the latest Python version correctly.

Resolves: #xxxxx

# Why this way


giacomo-alzetta-aiven:
The error for the lint step seems to be linked to `pylint` version and note any linting rules.
We need to specify a pylint version >= `2.16.0` to have python 3.11 support: https://pypi.org/project/pylint/2.16.0/
And then we may need to fix rule violations

aris-aiven:
By the way there's a warning for the `botocore-stubs[essential]` requirement, seems like we should either specify `boto3-stubs[essential]` (see https://github.com/youtype/mypy_boto3_builder#boto3) or simply `botocore-stubs` (see https://github.com/youtype/botocore-stubs)

aris-aiven:
Thinking out loud, pydantic should probably be constrained to `pydantic<2` in `setup.cfg` for the time being. Otherwise the latest version is installed (`pydantic==2.2.1` at the time of writing), which probably not what we want.

rominf (changes requested):
I was not quick enough and another PR was merged before yours, so you need to resolve conflicts.

requirements.dev.txt
```diff
@@ -5,7 +5,7 @@ mypy==1.5.1
 # Same version from pghoard
 pylint==2.17.4
 botocore
-botocore-stubs[essential]
+botocore-stubs
```
rominf, line 8: nit (non-blocking): this change is unrelated to pydantic, should probably go into separate commit

rohmu/util.py
```diff
@@ -18,7 +18,7 @@
 import platform
 import types
 
-LOG = logging.getLogger("rohmu.util")
+LOG = logging.getLogger(__name__)
```
rominf, line 21: nit (non-blocking): this change is unrelated to pydantic, should probably go into separate commit

rominf (approved):
LGTM: all the comments except the comment about pylint version comment are non-blocking; pylint version comment can be fixed later along with pylint upgrade.

aris-aiven:
Sorry @rominf the PR was a bit messy for sure

PR #128, opened 2023-07-17
multipart: proposal for concurrent API

giacomo-alzetta-aiven:

New API to perform multipart uploads allowing concurrent & non-monotonic upload of chunks.

# Why this way

Currently we do use multipart uploads to upload big files, but this is done serially by uploading one part at a time in monotonically increasing order. This limits concurrency.

In order to allow the clients to upload multiple parts concurrently we need to somehow expose the concept of a concurrent upload.

The API proposed here still hides a little bit of details.

# limits

As far as I can tell the limits for the different providers are:
* AWS S3: 10k chunks of at most 5 GiB
* GCP: 10k chunks of at most 5 GiB
* Azure: 50k chunks of at most 4GiB
* Swift: 1k+ chunks (depending on configuration) of at most 5GiB

I'm not sure if it makes much sense to do smart stuff to avoid the limits, since they are quite close and I'd expect the clients to be able to be configured for reducing the chunks size... 

I'd enforce by default at most 10k chunks and 4GiB per chunk since this should work for AWS/GCP/Azure... swift is configurable so doesn't make much sense to enforce that limit.

### DIFF
--- a/rohmu/object_storage/azure.py
+++ b/rohmu/object_storage/azure.py
@@ -66,7 +66,7 @@ class Config(StorageModel):
     proxy_info: Optional[ProxyInfo] = None
 
 
-class AzureTransfer(BaseTransfer[Config]):
+class AzureTransfer(BaseTransfer[Config]):  # pylint: disable=abstract-method
     config_model = Config
 
     def __init__(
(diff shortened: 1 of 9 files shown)


giacomo-alzetta-aiven:
Converted to draft since, after discussions, there is an alternative proposal which should probably be preferred.https://github.com/aiven/rohmu/pull/130

giacomo-alzetta-aiven:
Gone with the other approach

PR #132, opened 2023-08-18
tests: reorganize

rominf:

* tests: replace constants with fixtures
* `__init__.py` cleanup
* tests: make test and code hierarchies parallel

# Why this way

* `CONSTANT_TEST_RSA_PUBLIC_KEY` and `CONSTANT_TEST_RSA_PRIVATE_KEY` are not special enough to make them constants and require separate file for them. Replace them with fixtures for consistency.
* To be able to easily reason about top-level exported objects, make `__init__.py` clean. Note that to keep compatibility, everything present in the file before the change was re-exported.
* Test and code hierarchies have been mostly parallel, the change makes it consistent.


test/test_rohmu_dates.py
```diff
@@ -1,34 +0,0 @@
...
-See LICENSE for details
-"""
-from rohmu.dates import parse_timestamp
-
-import datetime
-import dateutil.tz
-import re
-
-
-def test_parse_timestamp() -> None:
```
akudiyar, line 14: Is this test now gone completely?

akudiyar (changes requested):
Looks like one test is missing

test/test_rohmu_dates.py
rominf:
```
$ git switch main
Already on 'main'
Your branch is up to date with 'origin/main'.
$ diff test/test_rohmu_dates.py test/test_dates.py && echo equal
equal
```

Probably I should have been written this explicitly in the commit message.

akudiyar (approved):

PR #133, opened 2023-08-19
object_storage: split providers into modules

rominf:

object_storage: split providers into modules
    
# Why this way

There is a need for reusing configuration models downstream, however
currently it is difficult, since configuration models are stored in the
same modules, as providers: configuration models imports cause
unnecessary imports such as provider libraries imports.

The solution is to split object storage providers into modules.

To keep compatibility, reimport objects, previously defined in provider
modules, in `__init__.py` files.


kmichel-aiven (changes requested):
The backward compatibility imports remove the intended benefits.

Because the `__init__.py` that does backward compatibility is a parent of each config module, importing a config module will still import the transfer module and its dependencies.

To get both backward compatibility and the fast imports, you need a separate hierarchy to store all config, outside of `rohmu/object_storage/providername`.

Since config models are quite small, and any use will probably need most models anyway,
a good option would be to have a single `rohmu/object_storage/config.py` file
with named models for each provider: `AzureObjectStorageConfig`, `GoogleObjectStorageConfig`, `LocalObjectStorageConfig`, etc...

And then have an alias import in their old place inside each provider folder, using the base `Config` name instead of the complete one.

(I'm pretty sure external users of the models would rather have differently named classe for each model instead of having all models called `Config` and having pick their own local alias for each one of them.)

rohmu/object_storage/azure.py
```diff
@@ -8,15 +8,26 @@
...
+    KEY_TYPE_OBJECT,
+    KEY_TYPE_PREFIX,
+    ProgressProportionCallbackType,
+)
+from rohmu.object_storage.config import (  # pylint: disable=unused-import
+    AZURE_MAX_BLOCK_SIZE as MAX_BLOCK_SIZE,
+    AzureObjectStorageConfig as Config,
+    calculate_azure_max_block_size as calculate_max_block_size,
+)
+from rohmu.typing import Metadata
```
rominf, line 30: Changes like these is there result of moving the lines back and forth with PyCharm refactoring tool. Absolute imports should be preferred, so I see no harm here. I can create a separate PR fixing all imports if needed.

kmichel-aiven (approved):

PR #134, opened 2023-08-22
compressor: replace snappy with cramjam

aris-aiven:

Replaces the [python-snappy](https://github.com/andrix/python-snappy/) dependency with [cramjam](https://github.com/milesgranger/pyrus-cramjam/tree/master/cramjam-python).

# Why this way

The python-snappy library is no longer maintained (see [1]). Cramjam is a drop-in replacement that allows for a considerable speedup both in the compression and decompression tasks. The relevant benchmark for snappy's framed format [2].

Before we accept this PR, we should benchmark cramjam ourselves with realistic workloads.

[1]: https://github.com/andrix/python-snappy/issues/124
[2]: https://github.com/milesgranger/pyrus-cramjam/blob/33c0516374fb9726ddcb82c5dfbe86be96d2cd35/cramjam-python/benchmarks/README.md

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,7 +70,6 @@ jobs:
 
       - id: dependencies
         run: |
-          sudo apt-get install -y libsnappy-dev
           pip install -r requirements.txt
           pip install -r requirements.dev.txt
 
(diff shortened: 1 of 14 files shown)


aris-aiven:
Not going to pursue this right now, we have higher priority work on our roadmap.

PR #137, opened 2023-08-24
misc: fix sphinx GitHub Actions workflow

rominf:

misc: fix sphinx GitHub Actions workflow

# Why this way

Write permissions were missing.


aris-aiven (approved):
👍

PR #138, opened 2023-08-25
Follow-up to address previous PR comments

aris-aiven:

- Add a comment regarding frozen requirements kept in sync with CI.
- Stop using deprecated `mock` package, migrate to `unittest.mock` instead

### DIFF
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -1,6 +1,5 @@
 # same versions than pghoard this must reviewed and pinned
 black==23.7.0
-mock
 mypy==1.5.1
 # Same version from pghoard
 pylint==2.17.4
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -1,8 +1,9 @@
-# same versions than pghoard this must reviewed and pinned
+# Kept in sync with the versions in CI
 black==23.7.0
 mypy==1.5.1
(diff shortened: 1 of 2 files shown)


requirements.dev.txt
```diff
@@ -1,9 +1,9 @@
-# same versions than pghoard this must reviewed and pinned
+# Kept in sync with the versions in CI
 black==23.7.0
-mock
 mypy==1.5.1
-# Same version from pghoard
 pylint==2.17.4
+isort==5.11.5
```
Mulugruntz, line 5: Should keep it alphabetically sorted.

Mulugruntz (commented):
Nice little improvement, thanks!
I left a small comment.

Also, about these version differences, maybe we could explode the reqs into several files
- `requirements.ci.txt`
- `requirements.dev.txt`
- `requirements.stubs.txt` (unsure about exploding this one there)

Might be easier to have dependabot to warn us on updates / security issues for the pinned ones.

Or also, not pin anything, so that it picks latest or the one that is required elsewhere in the deps graph.

rominf:
@Mulugruntz I would invest time into moving to modern Python packaging using `pyproject.toml` instead of splitting txt files :-)

kmichel-aiven (approved):

PR #139, opened 2023-08-27
Add support for bucket to bucket copy

kmichel-aiven:

This adds support for copy of multiple files across different buckets,
for now only if the two buckets are on the same provider.

# Why this way

This enables copying data from bucket to bucket without storing it locally.

### DIFF
--- a/rohmu/object_storage/google.py
+++ b/rohmu/object_storage/google.py
@@ -34,6 +34,7 @@ from rohmu.object_storage.base import (
     KEY_TYPE_OBJECT,
     KEY_TYPE_PREFIX,
     ProgressProportionCallbackType,
+    SourceStorageModelT,
 )
 from rohmu.object_storage.config import (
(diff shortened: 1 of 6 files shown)


rohmu/object_storage/google.py
```diff
@@ -299,7 +307,7 @@ def copy_file(
                 body=body,
                 destinationBucket=self.bucket_name,
                 destinationObject=destination_object,
-                sourceBucket=self.bucket_name,
+                sourceBucket=source_bucket,
```
alanfranz, line 310: shouldn't this be ```source_bucket.bucket_name``` ?

kmichel-aiven: It should :(

We really need some test buckets.

alanfranz (approved):
Thanks for the quick fix!

I just wonder why in copy_files_from we should accept a BaseTransfer and then reject non-GoogleTransfer instances, instead of just accepting only specific instances? (I often forget mypy covariance magic so it may be a moot point )

PR #140, opened 2023-08-27
Revert "object_storage: make object storage config field types more precise

kmichel-aiven:

This reverts commit 73fb83b6

# Why this way

This checks that the file or folder exists at the time the config object is created,
it's perfectly valid for the file or folder to not exist when the object is created
and sometimes impossible for that file or folder to exist:

The config may be created in a different environment from where the config
will be used in the end (different chroot, container or server).

Using the object type to validate a property that is actually not part of the
object but part of the external environment is flaky anyway: the environment
will change outside of the control of the object and the property validated
when the object was constructed will become false while the code using
the object will believe that property is still true.

### DIFF
--- a/rohmu/object_storage/config.py
+++ b/rohmu/object_storage/config.py
@@ -9,6 +9,7 @@ See LICENSE for details
 from __future__ import annotations
 
 from enum import Enum, unique
+from pathlib import Path
 from pydantic import Field
 from rohmu.common.models import ProxyInfo, StorageDriver, StorageModel
 from typing import Any, Dict, Final, Literal, Optional, TypeVar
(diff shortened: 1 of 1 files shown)


rominf (changes requested):
While I agree with the revert itself, it would be better to prevent this kind of mistakes in the future. I see two options: add `TypeAlias` for `str` with the explanation in the comment, or by adding comments in field definition lines for visibility. I would choose `TypeAlias`, since the knowledge that this is a path, but it is not validated, will be visible in IDEs without checking for comments.

kmichel-aiven:
> While I agree with the revert itself, it would be better to prevent this kind of mistakes in the future. I see two options: add `TypeAlias` for `str` with the explanation in the comment, or by adding comments in field definition lines for visibility. I would choose `TypeAlias`, since the knowledge that this is a path, but it is not validated, will be visible in IDEs without checking for comments.

Good point.
I've typed it as python's `Path`, which does not suffer from the same problem and added a comment.

rominf:
@kmichel-aiven Do we support something except POSIX (e.g. Windows)? I couldn't find this information in `setup.cfg`, hence the question. If we do, then this is probably incorrect, since on my Fedora machine:
```
In [1]: from pathlib import Path

In [2]: Path("C://")
Out[2]: PosixPath('C:')
```

kmichel-aiven:
> @kmichel-aiven Do we support something except POSIX (e.g. Windows)? I couldn't find this information in `setup.cfg`, hence the question. If we do, then this is probably incorrect, since on my Fedora machine:
> 
> ```
> In [1]: from pathlib import Path
> 
> In [2]: Path("C:\\")
> Out[2]: PosixPath('C:')
> ```

We don't support windows :)

rominf (approved):

PR #142, opened 2023-09-14
object_storage,s3: fix head_bucket call according to docs

rominf:

object_storage,s3: fix head_bucket call according to docs.

# Why this way

Previous method of error handling was wrong.

### DIFF
--- a/rohmu/object_storage/s3.py
+++ b/rohmu/object_storage/s3.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 
 from botocore.response import StreamingBody
 from functools import partial
+from http import HTTPStatus
 from rohmu.common.models import StorageOperation
 from rohmu.common.statsd import StatsdConfig
 from rohmu.errors import ConcurrentUploadError, FileNotFoundFromStorageError, InvalidConfigurationError, StorageError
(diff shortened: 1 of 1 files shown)


rikonen:
That 403 handling is intentional change from 2015 (https://github.com/Aiven-Open/pghoard/commit/d6c4cdfb702ec94785798ebb02af49e2aa950df5) and I suspect something will break for us due to the way how we set up the bucket policies if this is merged.

rohmu/object_storage/s3.py
```diff
@@ -542,12 +543,11 @@ def check_or_create_bucket(self) -> None:
...
             self.s3_client.head_bucket(Bucket=self.bucket_name)
         except botocore.exceptions.ClientError as ex:
+            # https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadBucket.html
             status_code = ex.response.get("ResponseMetadata", {}).get("HTTPStatusCode")
             if status_code == 301:
                 raise InvalidConfigurationError(f"Wrong region for bucket {self.bucket_name}, check configuration")
-            elif status_code == 403:
-                self.log.warning("Access denied on bucket check, assuming write permissions")
-            elif status_code == 404:
+            elif status_code in {HTTPStatus.BAD_REQUEST, HTTPStatus.FORBIDDEN, HTTPStatus.NOT_FOUND}:
```
giacomo-alzetta-aiven, line 550: Just to keep track of the discussion: in some places we rely on the current behaviour for 403/Forbidden of not creating a bucket and simply logging a message, so need to double check how to fix this part without breaking other stuff.

rominf:
It it good now?

giacomo-alzetta-aiven (approved):
LGTM

PR #143, opened 2023-09-19
fix Buffer issue after python version 3.10

Aiqin-Aiven:
pghoard is getting import Error for this Buffer in pipeline

Fix the import Error for python over 3.10 version
as Python 3.10 and later versions no longer provide 'Buffer' from 'typing_extensions'

# Why this way

still compatible with python 3.8 and 3.9


rominf (changes requested):
Duplicate (predecessor) of #144?

PR #144, opened 2023-09-20
make Buffer compatible with python 3.10 and above

Aiqin-Aiven:
Fix the error on pghoard pipeline for python 3.10 and above
E   ImportError: cannot import name 'Buffer' from 'typing_extensions' (/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/typing_extensions.py)

# Why this way
In Python 3.10 and later, the `Buffer` name is no longer provided by `typing_extensions`. Instead, you can use `typing.ByteString` or `typing.MutableSequence[bytes]` as a replacement for `Buffer`.

### DIFF
--- a/rohmu/util.py
+++ b/rohmu/util.py
@@ -10,7 +10,15 @@ from io import BytesIO, UnsupportedOperation
 from itertools import islice
 from rohmu.typing import HasFileno
 from typing import BinaryIO, Generator, Iterable, Optional, Tuple, TypeVar, Union
-from typing_extensions import Buffer
+
+import sys
+
+if sys.version_info < (3, 10):
+    from typing_extensions import Buffer
+else:
+    from typing import ByteString, MutableSequence
+
+    Buffer = Union[ByteString, MutableSequence[bytes]]
 
 import fcntl
 import logging
@@ -205,10 +213,10 @@ class ProgressStream(BinaryIO):
     def truncate(self, size: Optional[int] = None) -> int:
         raise UnsupportedOperation("truncate")
 
-    def write(self, s: Union[bytes, Buffer]) -> int:
+    def write(self, s: Union[bytes, Buffer]) -> int:  # type: ignore
         raise UnsupportedOperation("write")
 
-    def writelines(self, __lines: Union[Iterable[bytes], Iterable[Buffer]]) -> None:
+    def writelines(self, __lines: Union[Iterable[bytes], Iterable[Buffer]]) -> None:  # type: ignore
         raise UnsupportedOperation("writelines")
 
     def fileno(self) -> int:


rohmu/util.py
```diff
@@ -205,10 +213,10 @@ def seek(self, offset: int, whence: int = 0) -> int:
     def truncate(self, size: Optional[int] = None) -> int:
         raise UnsupportedOperation("truncate")
 
-    def write(self, s: Union[bytes, Buffer]) -> int:
+    def write(self, s: Union[bytes, Buffer]) -> int:  # type: ignore
```
aris-aiven, line 216: Why do you need to ignore the type here (and in `writelines`)?

giacomo-alzetta-aiven: `BinaryIO` wants only `bytes` and no `Buffer` as input for `write` since it's basically just `IO[bytes]`, but I believe in some places we actually do want to forward some buffer instead of `bytes` hence why `Buffer` appears here.

I believe that currently there is no way to avoid ignoring some cases because we have some dependencies that want `BinaryIO` and other places that don't want that in their typing. So if we remove `Buffer` here and make this a "proper" `BinaryIO` then you'll find similar comments when we use this class instead.

Aiqin-Aiven: should we go this way with type ignored? OR should I remove the Union of Buffer here? I tried locally, seems mypy didn't complain other places' typing error.

Aiqin-Aiven: in this repo: only 2 upload_concurent_chunk func calling this class: ProgressStream with fd: BinaryIO, seems all right.

Aiqin-Aiven: I believe we don't have other better option here, right?

giacomo-alzetta-aiven: Yeah I don't think we have better options now. Unfortunately this area of typing is a bit of a mess...

aris-aiven: The comment is not blocking from my side, was just trying to understand the issue. TY @giacomo-alzetta-aiven for breaking it down. The only concern was that it might hide legit issues that can pop up in the future. Maybe even that is overblown? I don't know.

rominf (changes requested):
Everything is much easier:
1. `Buffer` was introduced to `typing_extensions` in `4.6.0`: https://github.com/python/typing_extensions/blob/main/CHANGELOG.md#release-460-may-22-2023
2. `pghoard` pins `typing_extensions==4.5.0`: https://github.com/Aiven-Open/pghoard/blob/21bc21446977cadd2ac42e967bbba846d930f9d5/constraints.txt#L78
3. `typing_extensions.Buffer` supports all Python versions: https://github.com/python/typing_extensions/blob/c17c499b865585458bc334a1f895ebaedd2ab854/src/typing_extensions.py#L2522-L2548
4. So, you need just bump `typing_extensions` in `pghoard` to at least `4.6.0`. I would bump it to `4.7.1`, since it's what we currently have downstream.

Aiqin-Aiven:
As per: https://github.com/Aiven-Open/pghoard/pull/601
close this PR for now.

PR #145, opened 2023-09-26
s3: allow passing cert_path to S3Transfer

joelynch:
Allow passing a certificate to the s3 client.  This is needed for easy interop with MinIO.

Note that this is an option for botocore https://github.com/boto/botocore/blob/develop/botocore/session.py#L874

### DIFF
--- a/rohmu/object_storage/s3.py
+++ b/rohmu/object_storage/s3.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 from botocore.response import StreamingBody
 from functools import partial
 from http import HTTPStatus
+from pathlib import Path
 from rohmu.common.models import StorageOperation
 from rohmu.common.statsd import StatsdConfig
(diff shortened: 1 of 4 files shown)


rohmu/object_storage/s3.py
```diff
@@ -173,7 +175,7 @@ def __init__(
                 config=boto_config,
                 endpoint_url=custom_url,
                 region_name=region,
-                verify=is_verify_tls,
+                verify=str(cert_path) if cert_path is not None and is_verify_tls else is_verify_tls,
```
kmichel-aiven, line 178: Would be good to add tests for that, having a mistake in the verify flag would be quite catastrophic.

Also, I'd rather have an exception if `cert_path` is not `None` and  `is_verify_tls` is `False`.

Especially since the default value of `is_verify_tls` is `False`: it's too easy to write code that just specifies a `cert_path` parameter, and silently verifies nothing.

kmichel-aiven (dismissed):

kmichel-aiven (approved):

PR #146, opened 2023-10-04
Expose config models

joelynch:
Expose the Rohmu config models and add a method `get_transfer_from_model` to get a `transfer` from the pydantic model rather than having to serialize as dict first.  This will help with typing and hopefully eventually reduce the number of copies of these pydantic models currently in our codebase.

Other small improvements:
* Pylance / VSCode doesn't like importing the re-exports in `rohmu/__init__.py`, I think the addition of this `__all__` will change this.
* field `notifier` was confusing on a pydantic model since it is not itself a pydantic class.  Better off just passing this in directly to `BaseTransfer.from_model`
* Since `StorageModelT` is not covariant, it is _not_ the case that, for instance, `S3Transfer < BaseStorage[StorageModel]`.  Mypy correctly claims we cannot make this TypeVar covariant because, for example, if `transfer: S3Transfer` and `model: StorageModel` then we can't call `transfer.from_model(model)`.  Instead I have just typed using `Any` which will reduce the amount of `type: ignore`s downstream.  Using the new `get_transfer_from_model` has more narrow generic typing and should probably be used more downstream where possible.
* Ensure everything actually works in python3.8 since we are running that in CI ie don't use `|` for unions

### DIFF
--- a/rohmu/factory.py
+++ b/rohmu/factory.py
@@ -1,44 +1,48 @@
 # Copyright (c) 2023 Aiven, Helsinki, Finland. https://aiven.io/
 
-from .common.models import StorageModel
+from .common.models import StorageDriver, StorageModel
 from .errors import InvalidConfigurationError
 from .notifier.interface import Notifier
 from rohmu.object_storage.base import BaseTransfer
-from typing import Any, cast, Dict, Type
+from rohmu.object_storage.config import StorageModelT
+from typing import Any, Dict, Optional, Type
 
 STORAGE_TYPE = "storage_type"
 NOTIFIER_TYPE = "notifier_type"
 Config = Dict[str, Any]
 
 
-def get_class_for_transfer(obj_store: Config) -> Type[BaseTransfer[StorageModel]]:
-    storage_type = obj_store[STORAGE_TYPE]
-    if storage_type == "azure":
+def get_class_for_transfer(obj_store: Config) -> Type[BaseTransfer[Any]]:
+    return get_class_for_storage_driver(StorageDriver(obj_store[STORAGE_TYPE]))
+
+
+def get_class_for_storage_driver(storage_driver: StorageDriver) -> Type[BaseTransfer[Any]]:
+    if storage_driver == StorageDriver.azure:
         from rohmu.object_storage.azure import AzureTransfer
 
-        return cast(Type[BaseTransfer[StorageModel]], AzureTransfer)
-    elif storage_type == "google":
+        return AzureTransfer
+    elif storage_driver == StorageDriver.google:
         from rohmu.object_storage.google import GoogleTransfer
 
-        return cast(Type[BaseTransfer[StorageModel]], GoogleTransfer)
-    elif storage_type == "sftp":
+        return GoogleTransfer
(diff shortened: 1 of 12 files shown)


rohmu/factory.py
```diff
@@ -1,44 +1,48 @@
...
-    elif storage_type == "swift":
+        return S3Transfer
+    elif storage_driver == StorageDriver.swift:
         from rohmu.object_storage.swift import SwiftTransfer
 
-        return cast(Type[BaseTransfer[StorageModel]], SwiftTransfer)
+        return SwiftTransfer
 
-    raise InvalidConfigurationError(f"unsupported storage type {repr(storage_type)}")
+    raise InvalidConfigurationError(f"unsupported storage type {repr(storage_driver)}")
```
giacomo-alzetta-aiven, line 45: nit: could change `{repr(<expr>)}` with `{<expr>!r}`

rohmu/transfer_pool.py
```diff
@@ -137,7 +139,7 @@ def __getattribute__(self, attr: str) -> Any:
         return super().__getattribute__(attr)
 
     @classmethod
-    def from_model(cls, model: StorageModel) -> BaseTransfer[StorageModel]:
+    def from_model(cls, model: StorageModel, notifier: Notifier | None = None) -> Self:
```
giacomo-alzetta-aiven, line 142: Shouldn't we use `Optional` here?

giacomo-alzetta-aiven (changes requested):
I think this makes sense. But you missed a couple of places if we want to keep typing for 3.8.

rohmu/transfer_pool.py
joelynch: oops thanks

fingon (commented):
Doesn't that notifier change break existing usage? 

(I know it is broken, and I think we use it in only place but in general we try to avoid breaking backwards compatibility.)

I don't have better suggestions for it though. Perhaps we just have to upgrade the dependency at same time as rohmu.

giacomo-alzetta-aiven:
> Doesn't that notifier change break existing usage?

Yeah that's true. I guess we should probably bump rohmu version at least and make sure that change can be easily adapted to in our own code.

joelynch:
> Doesn't that notifier change break existing usage?

True if anyone was using `BaseTransfer.from_model` or `StorageModel.notifier`.  I don't think those methods were (until this PR) really supposed to be part of the "public api" for this library ie the stuff in `rohmu/__init__.py`.

I am not super attached to the `notifier` change but I think it is worthwhile, happy to revert if we want.

fingon:
> True if anyone was using `BaseTransfer.from_model` or `StorageModel.notifier`. I don't think those methods were (until this PR) really supposed to be part of the "public api" for this library ie the stuff in `rohmu/__init__.py`.

I'm pretty sure we at least have internal use (in btar) of that notifier.

> I am not super attached to the `notifier` change but I think it is worthwhile, happy to revert if we want.

I think the change makes sense, it is mostly question of how to roll it in. Perhaps we can just merge this in and then fix the dependencies as they crop up.

fingon (approved):

PR #150, opened 2023-10-11
Fix type hints

Mulugruntz:
This fixes some type hint issues that are preventing downstream projects from passing CI.

This contains the following changes:

- `SizeLimitedFile` uses common conventions, types and defaults.
- `_fileobj_name` -> `_obj_name`, as it's not related to files.
- `rohmufile.write_file` only requires `HasRead` for the input file.
- `HasRead`, `HasWrite` use Python conventions, types and defaults.
- `FileLike` uses Python conventions, types and defaults.
- New `HasName`.

# Why this way

### DIFF
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -21,3 +21,6 @@ types-mock
 types-requests
 # Extra stubs
 google-api-python-client-stubs
+# Used for their stubs
+zstandard==0.21.0
+python-snappy==0.6.1
--- a/rohmu/atomic_opener.py
+++ b/rohmu/atomic_opener.py
@@ -6,7 +6,7 @@ from typing_extensions import TypeAlias
 try:
     from typing import Literal
 except ImportError:
-    from typing_extensions import Literal  # type: ignore [assignment]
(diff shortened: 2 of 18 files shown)


fingon (changes requested):
If these whitespace changes are required by other projects, there should be autoformat step to ensure future rohmu changes also fullfill these requirements.

Mulugruntz:
> If these whitespace changes are required by other projects, there should be autoformat step to ensure future rohmu changes also fullfill these requirements. (And ideally also provides autofixed formatting if available.)

They are not required for the current problem this PR is fixing. If you like I can drop this particular commit that removes the spaces.

To be noted that the standard is **without** the space and that many tools will show a lot of alerts when that space is there.
https://mypy.readthedocs.io/en/stable/common_issues.html#spurious-errors-and-locally-silencing-the-checker
> You can use the form `# type: ignore[<code>]` to only ignore specific errors on the line.

fingon (approved):
It is fine to include it too. I don't mind the change, I just want to avoid hidden dependencies on stricter validation in other projects.

PR #154, opened 2023-10-13
Fix documentation

Mulugruntz:

Resolves: #xxxxx

# Why this way

I noticed the `docs/`, which seems to not be generated from our `README.rst`, even though it's a copy paste of it. (duplicates stuff).
That means that even though I fixed the links [in a previous PR](https://github.com/Aiven-Open/rohmu/pull/151), this was not reflected on [our docs website](https://aiven-open.github.io/rohmu/).
There was no process for that so far and it was just copy/pasting until now.

## To do

- [ ] change copyrights from docstrings to comments
- [ ] decide on a nice API page hierarchy.


Mulugruntz:
Continue on #155

PR #157, opened 2023-11-20
Do not throw exception if md5Hash, size or updated are missing from GCP

giacomo-alzetta-aiven:
list response

If `md5Hash` hash is missing from GCP list response use a hardcoded value instead.

If `size` is missing from GCP list response use `0` as hardcoded value instead.

If `updated` is missing from GCP list response use unix epoch as hardcoded value instead.

# Why this way

Apparently in some cases GCP does not provide the fields `md5Hash`, `size` and `updated`, so rohmu would raise a `KeyError` when trying to list such item.

There are 2 ways to handle this:
1. Make these field optional/accept `None` value at the API level for everyone. This is a major change. It would mean all users need to handle missing keys or the `None` value even if the backend their are using always provides these fields, and even for GCP it's extremely uncommon that they are missing
2. Use placeholder values to "fake" them.

This PR implements the second approach.

### DIFF
--- a/rohmu/object_storage/google.py
+++ b/rohmu/object_storage/google.py
@@ -46,6 +46,7 @@ from rohmu.util import get_total_size_from_content_range
 from typing import Any, BinaryIO, Callable, Collection, Iterable, Iterator, Optional, TextIO, Tuple, TypeVar, Union
 from typing_extensions import Protocol
 
+import base64
 import codecs
 import dataclasses
 import errno
@@ -168,6 +169,9 @@ class Reporter:
 
 
 ResType = TypeVar("ResType")
+
+# Seems like there are a few edge cases where MD5 field is missing when listing objects. Provide a fake value for those.
+_MISSING_MD5_B64_ENCODED = base64.encodebytes(b"Missing md5Hash!")
 
 
 class GoogleTransfer(BaseTransfer[Config]):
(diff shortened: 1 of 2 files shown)


giacomo-alzetta-aiven:
> Any idea when the values are missing? For any caller that actually relies on getting the values this will result in quite possibly incorrect behavior and that behavior may be worse than the hard fail that we'd get by making the values optional.

I'm thinking that maybe we should omit the keys if they are missing. In this way users can simply use `.get(<field>, default)` to handle it easily if in their use case a hardcoded value would work, or they can check for the key presence and handle it however they want.

Using `None` as value means they need to remember to do `.get(<field>) or default` and this is not even correct if their use case might have, e.g., 0-byte files.

I will open a different PR with that implementation.

giacomo-alzetta-aiven:
I guess we probably want to prefer this one instead: https://github.com/Aiven-Open/rohmu/pull/158

Keeping in mind that this seem to be a very rare occurrence, so 99.9999% of cases either version shouldn't change anything for the users. But if they happen to stumble on that extreme rare occurrence maybe failing explicitly is better.

giacomo-alzetta-aiven:
I guess we probably want to prefer this one instead: https://github.com/Aiven-Open/rohmu/pull/158

Keeping in mind that this seem to be a very rare occurrence, so 99.9999% of cases either version shouldn't change anything for the users. But if they happen to stumble on that extreme rare occurrence maybe failing explicitly is better.

PR #158, opened 2023-11-20
Handle rare case where GCP does not return size, updated and md5Hash

giacomo-alzetta-aiven:

If md5Hash hash is missing from GCP list response omit `md5` key from response

If size is missing from GCP list response omit `size` key from response.

If updated is missing from GCP list response omit `last_modfiied` from response.

# Why this way

Apparently in some cases GCP does not provide the fields md5Hash, size and updated, so rohmu would raise a KeyError when trying to list such item.

There are 2 ways to handle this:

1. Make these field optional, so all users should retriev them using `.get(<field>)` and handle the case where this returns `None`.
2. Use placeholder values to "fake" them.

This PR implements the first approach.

We omit the whole key so that users whose use case would work with a simple hardcoded value (e.g. size 0) can just do `.get("size", 0)`. If we used a `None` value they'd have to use `.get("size") or default` and this would not work if the object has 0 size (which can happen).


fingon (approved):
This approach seems fine to me. None would have been another option but in terms of typing that is quite hairy too.

PR #160, opened 2023-11-28
object_storage: avoid uploading empty file by multiparts

kathia-barahona:
When trying to upload an empty file by multipart for S3, botocore starts raising MalformedXML. In order to avoid this, rohmu should just verify if the file object is empty and just upload it entirely

Verifies if file object to be upload is empty, in this case multipart upload must not be used.

Resolves: #BF-2320

# Why this way

For S3, by default rohmu will always attempt multipart uploads (regardless if the content length is 0). For some storages this can be an issue, therefore we should just put directly the object into the storage. In case the file's metadata does not contain the content length and the file object is seekable, rohmu should just check if the file is indeed empty and always avoid multipart upload

### DIFF
--- a/rohmu/util.py
+++ b/rohmu/util.py
@@ -56,6 +56,19 @@ def set_stream_nonblocking(stream: HasFileno) -> None:
     fcntl.fcntl(fd, fcntl.F_SETFL, fl | os.O_NONBLOCK)
 
 
+def file_object_is_empty(fd: BinaryIO) -> bool:
+    # not seekable, so we cannot verify it has no contents
+    if not fd.seekable():
+        return False
+
+    fd.seek(0)
+    is_empty = len(fd.read(1)) == 0
+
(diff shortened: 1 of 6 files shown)


rohmu/util.py
```diff
@@ -56,6 +56,19 @@ def set_stream_nonblocking(stream: HasFileno) -> None:
     fcntl.fcntl(fd, fcntl.F_SETFL, fl | os.O_NONBLOCK)
 
 
+def file_object_is_empty(fd: BinaryIO) -> bool:
+    # not seekable, so we cannot verify it has no contents
+    if not fd.seekable():
+        return False
+
+    fd.seek(0, os.SEEK_END)
```
rdunklau, line 64: Can we just seek 1 byte instead of moving to the end ?

rdunklau (approved):

PR #162, opened 2023-12-07
object_storage: remove access denied warning for s3 buckets

tkren:

The `Access denied on bucket check` warning messages do not add value, at least they are confusing. A missing `s3:ListBucket` permission does not prohibit writing to a bucket.


rohmu/object_storage/s3.py
```diff
@@ -553,7 +553,8 @@ def check_or_create_bucket(self) -> None:
             if status_code == HTTPStatus.MOVED_PERMANENTLY:
                 raise InvalidConfigurationError(f"Wrong region for bucket {self.bucket_name}, check configuration")
             elif status_code == HTTPStatus.FORBIDDEN:
-                self.log.warning("Access denied on bucket check, assuming write permissions")
+                # Access denied on bucket check, most likely due to missing s3:ListBucket, assuming write permissions
```
aiven-amartin, line 556:
```suggestion
                # Access denied on bucket check, most likely due to missing s3:ListBucket, assuming we have write permissions
```

fingon (approved):

Issue #163, opened 2023-12-20
Support custom blob endpoint for Azurite


jeqo:
# What is currently missing?

trying to use rohmu on some local integration test with Azurite, I found that it's not possible to set a custom endpoint.

# How could this be improved?

Adding a host/post + disable HTTPS configurations to include a `BlobEndpoint` config into the configuration string as suggested here: https://learn.microsoft.com/en-us/azure/storage/blobs/use-azurite-to-run-automated-tests

# Is this a feature you would work on yourself?

* [x] I plan to open a pull request for this feature

PR #164, opened 2023-12-20
feat: add custom Azure host/port to support custom blob endpoint

jeqo:

Adds custom host and port configs to Azure to support custom blob endpoint like Azurite.

Resolves: #163 

# Why this way

Follow similar approach to S3, to set custom host/port.
Moved conn_string construction out of the initialization to be able to test it.

### DIFF
--- a/rohmu/object_storage/azure.py
+++ b/rohmu/object_storage/azure.py
@@ -24,6 +24,7 @@ from rohmu.object_storage.base import (
     SourceStorageModelT,
 )
 from rohmu.object_storage.config import (  # pylint: disable=unused-import
+    AZURE_ENDPOINT_SUFFIXES,
     AZURE_MAX_BLOCK_SIZE as MAX_BLOCK_SIZE,
     AzureObjectStorageConfig as Config,
     calculate_azure_max_block_size as calculate_max_block_size,
@@ -42,14 +43,6 @@ except ImportError:
     from azure.storage.blob._models import BlobPrefix, BlobType  # type: ignore
 
 
-ENDPOINT_SUFFIXES = {
-    None: "core.windows.net",
-    "germany": "core.cloudapi.de",  # Azure Germany is a completely separate cloud from the regular Azure Public cloud
-    "china": "core.chinacloudapi.cn",
-    "public": "core.windows.net",
-}
-
-
 # Reduce Azure logging verbocity of http requests and responses
 logging.getLogger("azure.core.pipeline.policies.http_logging_policy").setLevel(logging.WARNING)
 
@@ -64,6 +57,9 @@ class AzureTransfer(BaseTransfer[Config]):
         account_key: Optional[str] = None,
         sas_token: Optional[str] = None,
         prefix: Optional[str] = None,
+        is_secure: bool = True,
+        host: Optional[str] = None,
+        port: Optional[int] = None,
         azure_cloud: Optional[str] = None,
         proxy_info: Optional[dict[str, Union[str, int]]] = None,
         notifier: Optional[Notifier] = None,
@@ -78,16 +74,13 @@ class AzureTransfer(BaseTransfer[Config]):
         self.account_key = account_key
         self.container_name = bucket_name
         self.sas_token = sas_token
-        try:
-            endpoint_suffix = ENDPOINT_SUFFIXES[azure_cloud]
-        except KeyError:
-            raise InvalidConfigurationError(f"Unknown azure cloud {repr(azure_cloud)}")
-
-        conn_str = (
-            "DefaultEndpointsProtocol=https;"
-            f"AccountName={self.account_name};"
-            f"AccountKey={self.account_key};"
-            f"EndpointSuffix={endpoint_suffix}"
+        conn_str = self.conn_string(
+            account_name=account_name,
+            account_key=account_key,
+            azure_cloud=azure_cloud,
+            host=host,
+            port=port,
+            is_secure=is_secure,
         )
         config: dict[str, Any] = {"max_block_size": MAX_BLOCK_SIZE}
         if proxy_info:
@@ -97,13 +90,13 @@ class AzureTransfer(BaseTransfer[Config]):
                 auth = f"{username}:{password}@"
             else:
                 auth = ""
-            host = proxy_info["host"]
-            port = proxy_info["port"]
+            proxy_host = proxy_info["host"]
+            proxy_port = proxy_info["port"]
             if proxy_info.get("type") == "socks5":
                 schema = "socks5"
             else:
                 schema = "http"
-            config["proxies"] = {"https": f"{schema}://{auth}{host}:{port}"}
+            config["proxies"] = {"https": f"{schema}://{auth}{proxy_host}:{proxy_port}"}
 
         self.conn: BlobServiceClient = BlobServiceClient.from_connection_string(
             conn_str=conn_str,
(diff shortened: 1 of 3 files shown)


rohmu/object_storage/azure.py
```diff
@@ -113,6 +113,42 @@ def __init__(
...
+
+            conn_str = (
+                "DefaultEndpointsProtocol=https;"
+                f"AccountName={account_name};"
+                f"AccountKey={account_key};"
+                f"EndpointSuffix={endpoint_suffix}"
+            )
+        else:
+            print(host)
+            print(port)
```
rominf, line 139: Is it debugging leftovers?

rohmu/object_storage/azure.py
```diff
@@ -113,6 +113,42 @@ def __init__(
...
+        is_secure: bool,
+    ) -> str:
+        if not host and not port:
+            try:
+                endpoint_suffix = ENDPOINT_SUFFIXES[azure_cloud]
+            except KeyError:
+                raise InvalidConfigurationError(f"Unknown azure cloud {repr(azure_cloud)}")
+
+            conn_str = (
+                "DefaultEndpointsProtocol=https;"
```
rominf, line 132: Let's assume `is_secure == False`, but `host` and `port` are `None`. Returning connection string will be `https`, which is unexpected.

rohmu/object_storage/azure.py
```diff
@@ -113,6 +113,42 @@ def __init__(
...
+                raise InvalidConfigurationError("Custom host and port must be specified together")
+
+            protocol = "https" if is_secure else "http"
+            conn_str = (
+                f"DefaultEndpointsProtocol={protocol};"
+                f"AccountName={account_name};"
+                f"AccountKey={account_key};"
+                f"BlobEndpoint={protocol}://{host}:{port}/{account_name};"
+            )
+        return conn_str
```
rominf, line 150: There is duplication of 3/4 lines of `conn_str`. I think this can be rewritten to be readable and without duplication the same time.

test/object_storage/test_azure.py
```diff
@@ -103,3 +103,24 @@ def test_get_contents_to_fileobj_raises_error_on_invalid_byte_range(azure_module
             fileobj_to_store_to=BytesIO(),
             byte_range=(100, 10),
         )
+
+
+def test_conn_string_custom_host_port() -> None:
```
rominf, line 108: Since there is a strange behavior on I pointed out above, I suggest to write tests without custom host and port as well.

rominf (changes requested):

rohmu/object_storage/azure.py
jeqo: ofc 😅 , removed in last commit

rohmu/object_storage/azure.py
jeqo: Sure. Wasn't considering it for default endpoint, but there may be some scenario where this is needed. Refactored in last commit

rominf (approved):

exg (changes requested):
Fixup commits should be folded.

giacomo-alzetta-aiven:
Overall LGTM but I agree with Emanuele that commit history could be cleaned up before merging.

jeqo:
@exg @giacomo-alzetta-aiven thanks! Commits are folded now, have another look.

rohmu/object_storage/azure.py
```diff
@@ -113,6 +113,35 @@ def __init__(
...
+        if not host and not port:
+            try:
+                endpoint_suffix = ENDPOINT_SUFFIXES[azure_cloud]
+            except KeyError:
+                raise InvalidConfigurationError(f"Unknown azure cloud {repr(azure_cloud)}")
+
+            conn.append(f"EndpointSuffix={endpoint_suffix}")
+        else:
+            if not host or not port:
+                raise InvalidConfigurationError("Custom host and port must be specified together")
```
exg, line 140: It would be better to use pydantic validators for custom validation, see https://docs.pydantic.dev/1.10/usage/validators/

jeqo: Good idea. Adding this on the latest commit, have a look

exg: The validation of `azure_cloud` can also be moved in the validator. Also, remember to fold fixup commits before pushing :)

jeqo: Sure, can move it to validators as well.

Re: fixups, I prefer to keep them separate for easier review, but can surely rebase/squash once approved if that's ok.

exg: That is also fine, and I used to recommend that too before. I just find it unnecessary given that GitHub provides the diff between the previous and current branch tip on force push (via the "Compare" button).

giacomo-alzetta-aiven: Also if a reviewer marks the files as "viewed" when reviewing GitHub will also show them which files have changed when the PR is updated, so it's not like reviewers are forced to re-read everything to discover what you updated.

jeqo: Thanks, will keep that in mind. 
I've squashed all commits now for this change.

rohmu/object_storage/config.py
```diff
@@ -83,10 +89,28 @@ class AzureObjectStorageConfig(StorageModel):
...
+    host: Optional[str] = None
+    port: Optional[int] = None
     azure_cloud: Optional[str] = None
     proxy_info: Optional[ProxyInfo] = None
     storage_type: Literal[StorageDriver.azure] = StorageDriver.azure
 
+    @root_validator
+    @classmethod
+    def host_and_port_must_be_set_together(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        if "host" in values and "port" in values:
```
exg, line 102: This test is redundant because the fields have a default value and so they are always present. BTW, another option would be to introduce a `HostInfo` model, similar to `ProxyInfo`, containing the new fields. This way there would be no need for a custom validator, and it would be more consistent. WDYT?

jeqo: Good catch, fixing that on the latest commit. 

About HostInfo, I like the idea, but we could do that as a separate refactoring PR as there are many places where this type would fit. At the moment this is consistent with the S3 approach to set custom host/port.

exg: Agreed, I did not realize that this is based on the existing S3 config.

rohmu/object_storage/config.py
```diff
@@ -83,10 +89,27 @@ class AzureObjectStorageConfig(StorageModel):
...
+    host: Optional[str] = None
+    port: Optional[int] = None
     azure_cloud: Optional[str] = None
     proxy_info: Optional[ProxyInfo] = None
     storage_type: Literal[StorageDriver.azure] = StorageDriver.azure
 
+    @root_validator
+    @classmethod
+    def host_and_port_must_be_set_together(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        if not values["host"] or not values["port"]:
```
exg, line 102: This rejects (existing) configs where host and port are both missing, which is not correct. I think you want
```
(values["host"] is None) != (values["port"] is None)
```

giacomo-alzetta-aiven: Yeah, `test_azure_config_host_port_set_together` or a similar test should also verify that the valid cases do not raise an error.

jeqo: Oh, good catch! I added tests to validate the minimal case and fix this.

exg (approved):

rohmu/object_storage/azure.py
```diff
@@ -113,6 +106,28 @@ def __init__(
...
+        conn = [
+            f"DefaultEndpointsProtocol={protocol}",
+            f"AccountName={account_name}",
+            f"AccountKey={account_key}",
+        ]
+        if not host and not port:
+            endpoint_suffix = AZURE_ENDPOINT_SUFFIXES[azure_cloud]
+            conn.append(f"EndpointSuffix={endpoint_suffix}")
+        else:
+            conn.append(f"BlobEndpoint={protocol}://{host}:{port}/{account_name}")
```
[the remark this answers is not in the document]
exg: So it is possible to create transfers bypassing the pydantic config models?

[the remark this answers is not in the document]
giacomo-alzetta-aiven: I'm just saying that we are not indicating that `AzureTransfer` is private, so users could instantiate it directly instead of going through the factory methods that use `from_model` to create the instance from the pydantic configuration.

I don't this is a big deal, hence why I class this as nitpicking.

giacomo-alzetta-aiven:
Since we all agree let's merge this.

PR #165, opened 2023-12-22
encryptor: add support for symmetric encryption

exg:
This change adds variants of the Encryptor and Decryptor classes that use symmetric encryption.

### DIFF
--- a/rohmu/encryptor.py
+++ b/rohmu/encryptor.py
@@ -1,21 +1,22 @@
 """
 rohmu - content encryption
 
-Copyright (c) 2016 Ohmu Ltd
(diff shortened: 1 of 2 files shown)


rikonen (approved):
Looks good to me. Leaving open for other people to have a look.

For context, we need this for AR-479 (internal ticket) where we want to use (unique) AES keys directly without any intermediate RSA.

c-garcia (commented):
LGTM

exg:
@Aiven-Open/team-andromeda could you merge this PR? team-arch does have permission to merge at the moment.

PR #166, opened 2024-01-08
fix: keep endpoints from azure for compatibility

jeqo:
To maintain compatibility with old versions of the azure blob storage library


giacomo-alzetta-aiven (approved):
LGTM

Yeah should have caught this change in the original PR

PR #169, opened 2024-02-29
Switch to ruff for formatting and linting

exg:


Makefile
```diff
@@ -28,16 +28,15 @@ unittest:
...
 
 .PHONY: mypy
 mypy:
 	$(PYTHON) -m mypy --show-error-codes $(PYTHON_SOURCE_DIRS)
 
 .PHONY: fmt
 fmt:
-	isort $(PYTHON_SOURCE_DIRS)
-	black $(PYTHON_SOURCE_DIRS)
+	ruff format $(PYTHON_SOURCE_DIRS)
```
aiven-anton, line 39:
```suggestion
        ruff check --fix $(PYTHON_SOURCE_DIRS)
	ruff format $(PYTHON_SOURCE_DIRS)
```

suggestion: Also allow auto-fixing.

pyproject.toml
```diff
@@ -2,13 +2,31 @@
...
 
-[tool.isort]
-no_sections = true
-force_alphabetical_sort = true
-combine_as_imports = true
-profile = "black"
-skip_gitignore = true
-line_length = 125
+[tool.ruff.lint]
+select = [
```
aiven-anton, line 10:
```suggestion
extend-select = [
```

not (non-blocking): Prefer `extend-select` to not overwrite linters considered "sane defaults". Same applies to `extend-ignore`.

rohmu/atomic_opener.py
```diff
@@ -81,7 +81,7 @@ def _atomic_opener(
     _fd_spy(fd)
     try:
         file_obj = os.fdopen(fd, mode=mode, encoding=encoding)
-    except Exception:  # pylint: disable=broad-except
+    except Exception:
```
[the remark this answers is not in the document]
exg: I will add `BLE`, but note that it does not detect any of the existing broad except statements, because of https://github.com/astral-sh/ruff/issues/2212.

pyproject.toml
aiven-anton, line 10: suggestion: I would consider also enabling these. No opinion though as to whether they should be included in this initial introduction. (Copied from another project, with some amended comments).

```toml
extend-select = [
  # bugbear
  "B",
  # comprehensions
  "C4",
  # mccabe
  "C90",
  # bandit -- finds security issues, I ignore S101 when using this. Author of pre-commit summarizes
  # reasoning for that well here: https://stackoverflow.com/a/68429294/1220706.
  "S",
  # blind exception
  # Bare excepts are caught without this, but this also catches `except Exception: ...`.
  "BLE",
  # prevents shadowing builtins
  "A",
  # pycodestyle
  "W",
  # pyupgrade -- automatically upgrade to modern syntax
  "UP",
  # debugger
  "T10",
  # print -- might not be relevant
  "T20",
  # quotes
  "Q",
  # flake8-return
  # RET is theoretically nice, but there were too many false positives for it to be useful when I tried it.
  "RET",
  # simplify
  "SIM",
  # tidy imports
  # We use this to only outlaw relative parent imports.
  "TID",
]

[tool.ruff.mccabe]
max-complexity = 10

[tool.ruff.flake8-tidy-imports]
ban-relative-imports = "parents"
```

aiven-anton (approved):

rohmu/atomic_opener.py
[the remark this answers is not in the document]
aiven-anton: Ah, I wasn't aware of that nuance, that's disappointing. I made a comment here, explaining the use case of enforcing this strictly (regardless of logging and re-raising): https://github.com/astral-sh/ruff/issues/6421#issuecomment-1971255326

Makefile
exg: Added, limited to isort rules, as I don't think the `fmt` target should fix other violations.

pyproject.toml
exg: I prefer to start small, but I added some and fixed the associated violations.

Makefile
aiven-anton: That's OK, but now I'm curious as to why 😄 In my experience the ruff auto-fixers exceedingly rarely does the wrong thing.

pyproject.toml
```diff
@@ -2,13 +2,38 @@
...
+extend-ignore = [
+  "PLR0904", # too-many-public-methods
+  "PLR0912", # too-many-branches
+  "PLR0913", # too-many-arguments
+  "PLR0914", # too-many-locals
+  "PLR0915", # too-many-statements
+  "PLR0916", # too-many-boolean-expressions
+  "PLR2004", # magic-value-comparison
+  "UP006", # non-pep585-annotation
+  "UP007", # non-pep604-annotation
```
aiven-anton, line 31: question (non-blocking): Why not address these?

pyproject.toml
aiven-anton: Makes sense! 👍

pyproject.toml
exg: I prefer to fix those in a separate PR as some files use postponed evaluation of annotations while others don't, so one needs to be careful. Note that the minimum supported python version is 3.8, which predates PEP 585 and PEP 604.

Makefile
exg: Yes, it is not about that, but rather about semantics. I think that a target that is supposed to format code should not make unrelated changes.

aiven-anton (approved):

aiven-anton:
This LGTM, but I'm not authorized to merge in this repository.

giacomo-alzetta-aiven (approved):
LGTM

PR #173, opened 2024-03-01
feat: Add progress_callback to snapshot method

sebinsunny:
Updated the snapshot method to include a progress_callback for better tracking of operation. This allows for real-time progress tracking during snapshot operations, improving logging and debuggability for delta backup.

### DIFF
--- a/rohmu/delta/common.py
+++ b/rohmu/delta/common.py
@@ -4,13 +4,14 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 from datetime import datetime
+from enum import Enum
 from multiprocessing.dummy import Pool
 from pathlib import Path
 from pydantic import BaseModel, Field
 from rohmu.dates import now
 from rohmu.typing import AnyPath, HasRead
 from types import TracebackType
-from typing import Any, Callable, Iterable, List, Optional, Sequence, Type, TypeVar
+from typing import Any, Callable, Iterable, List, Optional, Sequence, Type, TypedDict, TypeVar
 
 import functools
 import hashlib
@@ -26,6 +27,20 @@ _log_1_1 = math.log(1.1)
 EMBEDDED_FILE_SIZE = 150
 
 logger = logging.getLogger(__name__)
+
+
+class ProgressMetrics(TypedDict, total=True):
+    handled: int
+    failed: int
+    total: int
+    final: bool
+
+
+class ProgressStep(Enum):
+    CREATING_MISSING_DIRECTORIES = "creating_missing_directories"
+    REMOVING_EXTRA_FILES = "removing_extra_files"
+    ADDING_MISSING_FILES = "adding_missing_files"
+    PROCESSING_AND_HASHING_SNAPSHOT_FILES = "processing_and_hashing_snapshot_files"
 
 
 def hash_hexdigest_readable(f: HasRead, *, read_buffer: int = 1_000_000) -> str:
@@ -253,6 +268,9 @@ class Progress(DeltaModel):
         if increase_worth_reporting(old_handled, self.handled, total=self.total):
             logger.debug("%s %r -> %r", info, n, self)
         assert not self.final
+
+    def progress_metrics(self) -> ProgressMetrics:
+        return {"handled": self.handled, "failed": self.failed, "total": self.total, "final": self.final}
 
     def download_success(self, size: int) -> None:
         self.add_success(size, info="download_success")
--- a/rohmu/delta/snapshot.py
+++ b/rohmu/delta/snapshot.py
@@ -10,6 +10,8 @@ from rohmu.delta.common import (
     increase_worth_reporting,
     parallel_map_to,
     Progress,
+    ProgressMetrics,
+    ProgressStep,
     SnapshotFile,
     SnapshotHash,
     SnapshotState,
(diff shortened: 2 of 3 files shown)


rohmu/delta/snapshot.py
```diff
@@ -218,7 +218,13 @@ def _snapshot_add_missing_files(self, *, src_files: Sequence[Path], dst_files: S
             changes += 1
         return changes
 
-    def snapshot(self, *, progress: Optional[Progress] = None, reuse_old_snapshotfiles: bool = True) -> int:
+    def snapshot(
+        self,
+        *,
+        progress: Optional[Progress] = None,
+        reuse_old_snapshotfiles: bool = True,
+        progress_callback: Optional[Callable[[str, bool], None]] = None,
```
RommelLayco, line 226: might be better to define a protocol for this

```python
class ProgressFuncProtocol(Protocol):
    def __call__(self, *, title: str, success: bool) -> None:
```

test/delta/test_snapshot.py
```diff
@@ -202,3 +202,51 @@ def fake_open_for_reading(self: SnapshotFile, path: Path) -> SizeLimitedFile:
...
             with pytest.raises(FileNotFoundError):
                 snapshotter.snapshot(progress=Progress(), reuse_old_snapshotfiles=True)
+
+
+@pytest.mark.timeout(2)
+def test_snapshot_without_create_samples(snapshotter_creator: Callable[..., SnapshotterWithDefaults]) -> None:
+    snapshotter = snapshotter_creator()
+    callback_messages: list[str] = []
+
+    def progress_callback(message: str, success: bool) -> None:
```
RommelLayco, line 212: when we use this function. will it need to be part of class so that it can append the messages to a field in the class?

If so we should define a protocol to ensure that all these methods are part of a class that contains this field

RommelLayco (changes requested):
How will these messages help us? Will something like pghoard read these message to determine what phase we are in, and how the phase succeeded? 

If so wouldn't it better to just pass this info into a dict instead of processing messages later?

test/delta/test_snapshot.py
sebinsunny: This is just an example for the test. progress_callback is used in PGHoard to monitor which phase we are in and whether it has succeeded.

rohmu/delta/common.py
```diff
@@ -254,6 +254,9 @@ def add_success(self, n: int = 1, *, info: str = "add_success") -> None:
             logger.debug("%s %r -> %r", info, n, self)
         assert not self.final
 
+    def progress_metrics(self) -> dict[str, int]:
```
nicois, line 257: it would be better to have a `TypedDict` as the return type (with `total=True`) as it will avoid incorrect assumptions about which keys are present

rohmu/delta/snapshot.py
```diff
@@ -258,14 +264,20 @@ def snapshot(self, *, progress: Optional[Progress] = None, reuse_old_snapshotfil
         # Create missing directories
         changes = self._snapshot_create_missing_directories(src_dirs=src_dirs, dst_dirs=dst_dirs)
         progress.add_success()
+        if progress_callback:
+            progress_callback("Creating missing directories", progress.progress_metrics())
 
         # Remove extra files
         changes += self._snapshot_remove_extra_files(src_files=src_files, dst_files=dst_files)
```
nicois, line 271: It might be better to have the progress callback called more frequently than this. I don't know in practice what the maximum time it might be for this method to complete, but I imagine for services which experience a lot of flux, it could be high.
This could be coupled with a debouncer to avoid calling it frequently, during these loops.

rohmu/delta/common.py
sebinsunny: included TypedDict for progress_metrics

rohmu/delta/snapshot.py
sebinsunny: yep, added callback for _snapshot_remove_extra_files

nicois:
This looks reasonable to me.

It would be nicer to have a debounced call to the progress callback inside each loop, but not essential for this iteration; it can always be added as a subsequent PR if need be.

rohmu/delta/snapshot.py
```diff
@@ -177,13 +178,22 @@ def _snapshot_create_missing_directories(self, *, src_dirs: Sequence[Path], dst_
             changes += 1
         return changes
 
-    def _snapshot_remove_extra_files(self, *, src_files: Sequence[Path], dst_files: Sequence[Path]) -> int:
+    def _snapshot_remove_extra_files(
+        self,
+        *,
+        src_files: Sequence[Path],
+        dst_files: Sequence[Path],
+        progress: Progress,
```
nicois, line 186: if `progress` is required, this will be a backwards-compatible change. It may be better to have a(n immutable) default value here.

sebinsunny: yep, I made the progress parameter optional and default to None. Inside the method, if progress isn't provided, a new Progress instance is created, ensuring backward compatibility and optional progress tracking.

RommelLayco (approved):
lgtm. One nit could be to make the `progress_callback` type use mypy protocol this is to ensure that the protocol passes in a method that references some object that can be used to read the callback

facetoe (approved):

PR #175, opened 2024-03-26
Use S3 dualstack endpoint by default

hnousiainen:
When using the default S3 endpoints, request using dualstack IPv4/IPv6 endpoints.

Configure botocore to utilize dualstack IPv4/IPv6 endpoint for S3. This enables IPv6 only clients to use S3 through rohmu as well.

### DIFF
--- a/rohmu/object_storage/s3.py
+++ b/rohmu/object_storage/s3.py
@@ -120,6 +120,7 @@ class S3Transfer(BaseTransfer[Config]):
         read_timeout: Optional[float] = None,
         notifier: Optional[Notifier] = None,
         aws_session_token: Optional[str] = None,
+        use_dualstack_endpoint: Optional[bool] = True,
         statsd_info: Optional[StatsdConfig] = None,
     ) -> None:
         super().__init__(prefix=prefix, notifier=notifier, statsd_info=statsd_info)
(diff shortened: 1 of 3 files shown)


rohmu/object_storage/s3.py
```diff
@@ -137,6 +137,7 @@ def __init__(
             if proxy_info:
                 proxy_url = get_proxy_url(proxy_info)
                 custom_config["proxies"] = {"https": proxy_url}
+            custom_config["use_dualstack_endpoint"] = True
```
rikonen, line 140: Should this be provided as a parameter so that calling code is able to control the value in case this ends up causing some problems?

hnousiainen: I added a configuration option to turn this off, but I'd personally prefer keeping it enabled by default: IPv6 is nowdays well supported, if available and configured.

rikonen: That actually reminds me, was this validated against non-AWS implementations of S3? I at least have no idea what's the state of IPv6 with those.

hnousiainen: The config only applies to default botocore addresses, and is not used if one passes explicit host/port information.

rikonen (approved):

PR #180, opened 2024-04-17
azure: allow StrEnum use in AzureTransfer.get_or_create_container

joelynch:
Due to how the Azure client works [here](https://github.com/Azure/azure-sdk-for-python/blob/azure-storage-blob_12.8.1/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_service_client.py#L667) using a subclass of Enum for the `container_name` with AzureTransfer leads to confusing behaviour - the enum _key_ is used rather than the enum _value_ as would be expected. This fixes that behaviour.


giacomo-alzetta-aiven (approved):
LGTM

PR #183, opened 2024-05-01
replace dateutil with fromisoformat in google backend

kmichel-aiven:

replace dateutil with fromisoformat in google backend.

# Why this way

`dateutil.parser.parse` is particularly slow and covers a lot of use cases that are not relevant to parsing [RFC 3339 datetimes](https://googleapis.github.io/google-api-python-client/docs/dyn/storage_v1.objects.html#list).

This was bad enough that rohmu could eat >50% of CPU while listing entries from Google object storage.

(The next high CPU user is `json.decode`, but that's still far behind `dateutil` and harder to fix since it's inside `googleapiclient`.)

dateutil_slow


aris-aiven (approved):
lgtm!

PR #184, opened 2024-05-02
Use hatch as build backend

joelynch:
Use hatch as a build backend.

Other small changes:
* Remove unused rpm spec file.
* Version file is now generated by setuptools_scm - for this to work,
  needed to change the format of the release tags.
* In release script, do not modify untracked version file.


kmichel-aiven (approved):
This was double-checked with a downstream build, so LGTM.

PR #187, opened 2024-07-18
Close google/azure/s3 client when closing GoogleTransfer object [DDB-1160]

kmichel-aiven:

GoogleTransfer objects create clients and never close the client, if we create many clients, we leak sockets.

Add a API to close a Transfer object and implement it in the GoogleTransfer subclass to close the google client.

Also implement it in Azure/S3 transfer objects for consistency (and hopefully one day we'll stop creating buckets during `__init__`, which will make the objects creation cheaper and free of side-effects.)

# Why this way

Because if we leak socket, then we eventually crash.

### DIFF
--- a/rohmu/object_storage/base.py
+++ b/rohmu/object_storage/base.py
@@ -82,6 +82,10 @@ class BaseTransfer(Generic[StorageModelT]):
         self.notifier = notifier or NullNotifier()
         self.stats = StatsClient(statsd_info)
 
+    def close(self) -> None:
+        """Release all resources associated with the Transfer object."""
+        pass
+
     @staticmethod
     def _incremental_to_proportional_progress(
         *, size: int, cb: ProgressProportionCallbackType
--- a/rohmu/object_storage/google.py
+++ b/rohmu/object_storage/google.py
@@ -188,9 +188,17 @@ class GoogleTransfer(BaseTransfer[Config]):
         self.proxy_info = proxy_info
         self.google_creds = get_credentials(credential_file=credential_file, credentials=credentials)
         self.gs: Optional[Resource] = self._init_google_client()
-        self.gs_object_client = None
+        self.gs_object_client: Any = None
         self.bucket_name = self.get_or_create_bucket(bucket_name)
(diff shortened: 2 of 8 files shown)


rohmu/object_storage/base.py
```diff
@@ -82,6 +82,9 @@ def __init__(
         self.notifier = notifier or NullNotifier()
         self.stats = StatsClient(statsd_info)
 
+    def close(self) -> None:
```
Khatskevich, line 85: I do not understand who is calling this close call?
I could not find api in this repo which will enforce the `close` call.
Should we then add `close` call to the destructor to enforce at least postponed resource cleanup?

Khatskevich: Otherwise a client pool may be used. Or any other simple way which makes it easy to use and reliable

rohmu/object_storage/google.py
```diff
@@ -188,10 +188,18 @@ def __init__(
         self.proxy_info = proxy_info
         self.google_creds = get_credentials(credential_file=credential_file, credentials=credentials)
         self.gs: Optional[Resource] = self._init_google_client()
-        self.gs_object_client = None
+        self.gs_object_client: Optional[Any] = None
```
joelynch, line 191: nit:
```suggestion
        self.gs_object_client: Any = None
```

joelynch (approved):
Seems good to me.

rohmu/object_storage/base.py
joelynch: Close is part of the API for this class.  A docstring might help here.

kmichel-aiven: Files, sockets, etc... need to be closed to reliably not have leaks, the same is true for APIs that use them.

I'll add a docstring to make it clear that this is part of the API.

rohmu/object_storage/google.py
kmichel-aiven: Fixed.

joelynch (approved):
LGTM

PR #188, opened 2024-07-23
Fix release tooling not working

kmichel-aiven:

Fix the tooling to be able to make a release.

# Why this way

Because we can't make a release if the tools are broken.


alanfranz (approved):
LGTM

PR #191, opened 2024-08-15
Fix Azure typing issues

jadkik:

The latest version of the azure-storage-blob package seems to have introduced some typing issues (most likely via https://github.com/Azure/azure-sdk-for-python/commit/cb47de53f3716757448343a2603e2f1e3feddc6d).

This PR solves this.

Resolves: #xxxxx

# Why this way

The version of `azure-storage-blob` is explicitly un-pinned in the `pyproject.toml` file, so it seems like a better option to address the typing issue rather than pin the package to an older version.

I opted for a warning log line instead of an assertion as that would at least provide some insight into what is happening. Theoretically, we might get in this situation even though the docs say we won't.

### DIFF
--- a/rohmu/object_storage/azure.py
+++ b/rohmu/object_storage/azure.py
@@ -168,7 +168,16 @@ class AzureTransfer(BaseTransfer[Config]):
                 if time.monotonic() - start < timeout:
                     time.sleep(0.1)
                 else:
-                    destination_client.abort_copy(copy_props.id, timeout=timeout)
+                    if copy_props.id is None:
+                        # The description of CopyProperties tells us copy_id should not be None unless status is also None
+                        # The type checker cannot know this, but we still log a warning if this ever happens
+                        self.log.warning(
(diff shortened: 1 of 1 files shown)


rohmu/object_storage/azure.py
```diff
@@ -168,7 +168,16 @@ def _copy_file_from_bucket(
                 if time.monotonic() - start < timeout:
                     time.sleep(0.1)
                 else:
-                    destination_client.abort_copy(copy_props.id, timeout=timeout)
+                    if copy_props.id is None:
+                        # The description of CopyProperties tells us copy_id should not be None unless status is also None
+                        # The type checker cannot know this, but we still log a warning if this ever happens
+                        self.log.warning(
+                            "Pending copy operation from %r to %r was missing a copy_id, will not be aborted after timeout",
```
bugant, line 175: nit: "will not be aborted after timeout" -> "thus cannot be aborted"

bugant (approved):

PR #192, opened 2024-08-21
Add progress callback for transferred object keys in copy_files_from

tilman-aiven:

This adds a callback function in `copy_files_from`, which provides the transferred object key with completed and total files.

We want to measure the progress of transferred files. However measuring the amount of transferred bytes requires a provider based implementation. Therefore we measure the keys which is provider independent.

### DIFF
--- a/rohmu/object_storage/base.py
+++ b/rohmu/object_storage/base.py
@@ -56,6 +56,10 @@ ProgressProportionCallbackType = Optional[Callable[[int, int], None]]
 IncrementalProgressCallbackType = Optional[Callable[[int], None]]
 
 
+class ObjectTransferProgressCallback(Protocol):
+    def __call__(self, files_completed: int, total_files: int) -> None: ...
+
+
 @dataclass(frozen=True, unsafe_hash=True)
 class ConcurrentUpload:
     backend: str
@@ -202,10 +206,19 @@ class BaseTransfer(Generic[StorageModelT]):
         cannot be copied with this method. If no metadata is given copies the existing metadata."""
         raise NotImplementedError
 
-    def copy_files_from(self, *, source: BaseTransfer[SourceStorageModelT], keys: Collection[str]) -> None:
+    def copy_files_from(
+        self,
+        *,
+        source: BaseTransfer[Any],
+        keys: Collection[str],
+        progress_fn: ObjectTransferProgressCallback | None = None,
+    ) -> None:
         if isinstance(source, self.__class__):
-            for key in keys:
+            total_files = len(keys)
+            for index, key in enumerate(keys):
                 self._copy_file_from_bucket(source_bucket=source, source_key=key, destination_key=key, timeout=15)
+                if progress_fn is not None:
+                    progress_fn(index + 1, total_files)
         else:
             raise NotImplementedError
 
--- a/rohmu/object_storage/base.py
+++ b/rohmu/object_storage/base.py
@@ -203,6 +203,21 @@ class BaseTransfer(Generic[StorageModelT]):
         raise NotImplementedError
 
     def copy_files_from(self, *, source: BaseTransfer[SourceStorageModelT], keys: Collection[str]) -> None:
+        if isinstance(source, self.__class__):
+            for key in keys:
+                self._copy_file_from_bucket(source_bucket=source, source_key=key, destination_key=key, timeout=15)
+        else:
+            raise NotImplementedError
+
+    def _copy_file_from_bucket(
+        self,
+        *,
+        source_bucket: Self,
+        source_key: str,
+        destination_key: str,
+        metadata: Optional[Metadata] = None,
+        timeout: float = 15.0,
+    ) -> None:
         raise NotImplementedError
 
     def format_key_for_backend(self, key: str, remove_slash_prefix: bool = False, trailing_slash: bool = False) -> str:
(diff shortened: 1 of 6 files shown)


test/object_storage/test_object_storage.py
```diff
@@ -72,15 +74,15 @@ def test_copy(transfer_type: str, request: Any) -> None:
...
     source.store_file_from_memory("some/a/key.ext", b"content_a", metadata={"info": "aaa"})
     source.store_file_from_memory("some/b/key.ext", b"content_b", metadata={"info": "bbb"})
-    destination.copy_files_from(
-        source=source,
-        keys=["some/a/key.ext", "some/b/key.ext"],
-    )
+    destination.copy_files_from(source=source, keys=["some/a/key.ext", "some/b/key.ext"], progress_fn=mock_progress_fn)
     assert destination.get_contents_to_string("some/a/key.ext") == (b"content_a", {"info": "aaa", "Content-Length": "9"})
     assert destination.get_contents_to_string("some/b/key.ext") == (b"content_b", {"info": "bbb", "Content-Length": "9"})
+    assert mock_progress_fn.call_count == 2
```
aris-aiven, line 84: Why not test the actual keys were called here?

rohmu/object_storage/base.py
```diff
@@ -55,6 +55,9 @@ class IterKeyItem(NamedTuple):
 # Argument is the additional number of bytes transferred
 IncrementalProgressCallbackType = Optional[Callable[[int], None]]
 
+# Argument is the transferred object key
+ObjectTransferProgressCallbackType = Optional[Callable[[str], None]]
```
aris-aiven, line 59: nit: could use a `TypeAlias` annotation
```suggestion
from typing import TypeAlias
ObjectTransferProgressCallbackType: TypeAlias = Optional[Callable[[str], None]]
```

aris-aiven (commented):
Small comments, overall looks good

test/object_storage/test_object_storage.py
tilman-aiven: Replaced with simply
```python
assert mock_progress_fn.call_args_list == [mock.call("some/a/key.ext"), mock.call("some/b/key.ext")]
```

Previous assertion  `mock_progress_fn.assert_has_calls` did not raise when the call arg list included additional calls.

rohmu/object_storage/base.py
tilman-aiven: `TypeAlias` is introduced in Python 3.10+, however we support 3.8+. Probably thats why we haven' t used it.
Added `TypeAlias` import from `typing_extensions`.

rohmu/object_storage/base.py
```diff
@@ -202,7 +205,30 @@ def copy_file(
...
+        *,
+        source: BaseTransfer[SourceStorageModelT],
+        keys: Collection[str],
+        progress_fn: ObjectTransferProgressCallbackType = None,
+    ) -> None:
+        if isinstance(source, self.__class__):
+            for key in keys:
+                self._copy_file_from_bucket(source_bucket=source, source_key=key, destination_key=key, timeout=15)
+                if progress_fn is not None:
+                    progress_fn(key)
```
kmichel-aiven, line 219: I don't think the progress callback should take the key name as the parameter :

That prevents backward compatible implementation changes/optimization where the progress might jump suddenly by multiple files (if we do batches, etc.) and it would be silly to call the callback 100 times after a batch of 100 uploads, just to have the callback increment a counter by 1, 100 times.

I think a `progress_fn(completed_files: int, total_files: int)` is a better interface here, and happens to be more similar to what we have in other places for bytes, and more aligned with the needs of the library users.

kmichel-aiven (dismissed):

rohmu/object_storage/base.py
tilman-aiven: `progress_fn(completed_files: int, total_files: int)` is a better interface in terms of counting. The case that that we might skip or jump (bulk) a transfer is valid.

Initial idea
- provide more info about the actual transferred object
- but let the caller calculate the metrics

rohmu/object_storage/base.py
```diff
@@ -202,7 +205,31 @@ def copy_file(
         cannot be copied with this method. If no metadata is given copies the existing metadata."""
         raise NotImplementedError
 
-    def copy_files_from(self, *, source: BaseTransfer[SourceStorageModelT], keys: Collection[str]) -> None:
+    def copy_files_from(
+        self,
+        *,
+        source: BaseTransfer[SourceStorageModelT],
```
joelynch, line 211: nit: having a generic appear only once in a function signature doesn't really do anything?  I am not convinced that this has an advantage over `BaseTransfer[Any]`.

joelynch (approved):
Few small typing style nits but otherwise looks great

rohmu/object_storage/base.py
```diff
@@ -55,6 +55,9 @@ class IterKeyItem(NamedTuple):
 # Argument is the additional number of bytes transferred
 IncrementalProgressCallbackType = Optional[Callable[[int], None]]
 
+# First argument is the number of transferred objects, second is the total number of objects
+ObjectTransferProgressCallbackType: TypeAlias = Optional[Callable[[int, int], None]]
```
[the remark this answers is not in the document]
tilman-aiven: Followed suggestion from [comment](https://github.com/Aiven-Open/rohmu/pull/192#discussion_r1725138490).
We are importing the type alias in astacus. 
The initial approach was targeting consistency with the previous callbacks.

It makes sense to let the implementer decide, if the callback is optional or not 👍 
Having named arguments in `Protocol` is pretty useful 👍

aris-aiven (approved):
Comments addressed, thank you for the contribution

PR #194, opened 2024-10-28
Re-reaise IncompleteReadError as MaybeRecoverableError

orange-kao:

S3 and Azure can throw IncompleteReadError when the TCP session was disrupted. However, current implementation does not re-reaise it as MaybeRecoverableError. As a result, programs like pghoard will give up basebackup download.

This is more likely to affect large (2~4 TiB) databases, because large database takes longer to download base backup and more likely to encounter at least one IncompleteReadError.

This commit re-reaise IncompleteReadError as MaybeRecoverableError, so pghoard can retry on an object without starting over.

# Why this way

I learned that pghoard use `MaybeRecoverableError` to decide whether it will retry or not. So I thought re-reaise IncompleteReadError as MaybeRecoverableError might be an option.

### DIFF
--- a/rohmu/object_storage/azure.py
+++ b/rohmu/object_storage/azure.py
@@ -5,12 +5,13 @@
 
 from __future__ import annotations
 
-from azure.core.exceptions import HttpResponseError, ResourceExistsError
+from azure.core.exceptions import HttpResponseError, IncompleteReadError, ResourceExistsError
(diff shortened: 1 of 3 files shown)


orange-kao:
I have force-pushed the branch after fixing for `make lint`.

aris-aiven (approved):
👍

PR #197, opened 2025-01-15
Add option to preserve trailing slash on deletes

giacomo-alzetta-aiven:
Most object storages support having a slash in the object
name and rohmu should be able to delete such objects.

Let the default behaviour the same but introduce an option
`preserve_trailing_slash` which allows to control whether the trailing
slash in the key are stripped or not.

Resolves: #198 

# Why this way

Introducing a new flag is backward compatible and preserves the existing behaviour in case this is actually a requirement for some user.

We can later on switch the default value of `preserve_trailing_slash` to `True` if we think we want this behaviour by default.

For local/sftp which can't support slashes I opted for explicitly raising an `Error` if the user passes `preserve_trailing_slash=True` instead of just ignoring the parameter.

### DIFF
--- a/test/object_storage/test_local.py
+++ b/test/object_storage/test_local.py
@@ -3,12 +3,14 @@ from concurrent.futures import ThreadPoolExecutor
 from functools import partial
 from io import BytesIO
 from itertools import cycle
-from rohmu.errors import FileNotFoundFromStorageError, InvalidByteRangeError
+from rohmu.errors import Error, FileNotFoundFromStorageError, InvalidByteRangeError
 from rohmu.object_storage.base import KEY_TYPE_OBJECT
 from rohmu.object_storage.local import LocalTransfer
 from tempfile import NamedTemporaryFile, TemporaryDirectory
+from typing import Union
 from unittest.mock import MagicMock
 
+import glob
 import hashlib
 import json
 import os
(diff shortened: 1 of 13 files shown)


test/object_storage/test_local.py
```diff
@@ -297,3 +299,71 @@ def inc_progress(size: int) -> None:
...
+            notifier=notifier,
+            prefix="test-prefix/",
+        )
+
+        transfer.store_file_from_memory("2", memstring=b"Hello")
+        transfer.store_file_from_memory("3", memstring=b"Hello")
+        transfer.store_file_from_memory("4/", memstring=b"Hello")
+        found_files = glob.glob(os.path.join(destdir, "test-prefix", "*"))
+        # ensure we have created some files
+        assert found_files
```
jadkik, line 357: nit: should we assert on the actual files that were created? Like `list(found_files) == ["2", "3", "4"]` but with full paths perhaps

Since we really want to test that the file `4` is created and then deleted.

jadkik (approved):
LGTM, really minor comment on one of the tests

test/object_storage/test_local.py
giacomo-alzetta-aiven: There is a separate test for `store_file_from_memory` so in theory we could just remove this assertion and trust that the files are created as expected.

bugant (approved):

PR #200, opened 2025-02-19
Fix Azure empty object download

ilpon:

Fix error handling to catch 416 response from Azure with additional check for blob size to map it to empty object.

# Why this way

The check for 416 HTTP status code already existed under wrong exception catch and only needed to be moved to under correct exception.

As the error only indicates that the range is invalid, an additional check for the blob size is added.


jadkik (approved):
LGTM

giacomo-alzetta-aiven (approved):
LGTM!

Nice finding!

PR #201, opened 2025-02-26
google: retry on httplib2.ServerNotFoundError

bhammerich-aiven:

Resolves: Errors of ServerNotFound wihle executing a request.

# Why this way

### DIFF
--- a/rohmu/object_storage/google.py
+++ b/rohmu/object_storage/google.py
@@ -308,12 +308,22 @@ class GoogleTransfer(BaseTransfer[Config]):
         while True:
             try:
                 return action()
-            except (IncompleteRead, HttpError, ssl.SSLEOFError, socket.timeout, OSError, socket.gaierror) as ex:
+            except (
+                IncompleteRead,
+                HttpError,
+                ssl.SSLEOFError,
+                socket.timeout,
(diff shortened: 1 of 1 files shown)


rohmu/object_storage/google.py
jadkik, line 1: note: There's some redundancy which seems a bit odd:

https://github.com/Aiven-Open/rohmu/blob/31bfb06f2f33d1c1a1f6f716e7c58b78a9370edb/rohmu/object_storage/google.py#L260-L266

We are supposed to handle it here, and it has its own retry logic.

And then in this function we seem to handle another kind of DNS error:

https://github.com/Aiven-Open/rohmu/blob/454acb663aaede1063fcd6a003d68f0ee8991afb/rohmu/object_storage/google.py#L337-L339

Nothing to do here, just an observation :smile:

jadkik (approved):
LGTM, just left a note but I don't think we need to change/refactor anything in here.

rohmu/object_storage/google.py
bhammerich-aiven: yeah it seems like googleapi sometimes fails unexpectedly

bugant (approved):

PR #203, opened 2025-04-18
s3: respect max num of parts per upload

ali-gholami-aiven:
maximum number of parts per upload for s3 compatible api is 10_000.

make sure we respect that and if needed we increase the chunk size to avoid starting a multipart upload with number of chunks (parts) that we already know is not valid.

chunk size is increased by multiplying previous size by 2 until we reach maximum chunk size.

### DIFF
--- a/rohmu/object_storage/s3.py
+++ b/rohmu/object_storage/s3.py
@@ -33,7 +33,9 @@ from rohmu.object_storage.base import (
 )
 from rohmu.object_storage.config import (  # noqa: F401
     calculate_s3_chunk_size as calculate_chunk_size,
-    S3_MULTIPART_CHUNK_SIZE as MULTIPART_CHUNK_SIZE,
+    S3_DEFAULT_MULTIPART_CHUNK_SIZE as MULTIPART_CHUNK_SIZE,
+    S3_MAX_NUM_PARTS_PER_UPLOAD,
+    S3_MAX_PART_SIZE_BYTES,
     S3_READ_BLOCK_SIZE as READ_BLOCK_SIZE,
     S3AddressingStyle,
     S3ObjectStorageConfig as Config,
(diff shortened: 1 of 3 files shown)


rohmu/object_storage/s3.py
```diff
@@ -486,6 +488,26 @@ def get_file_size(self, key: str) -> int:
...
+        """
+        if size is None:
+            return 1, self.default_multipart_chunk_size
+        chunks = math.ceil(size / self.default_multipart_chunk_size)
+        chunk_size = self.default_multipart_chunk_size
+
+        while chunks > S3_MAX_NUM_PARTS_PER_UPLOAD:
+            chunk_size *= 2
+            if chunk_size > S3_MAX_PART_SIZE_BYTES:
+                raise StorageError(f"Cannot upload a file of size {size}. Chunk size {chunk_size} is too big for each part of multipart upload.")
```
ali-gholami-aiven, line 506: not sure if raising `StorageError` is the best thing here as it might confuse user that this is an issue on storage side. please let me know if you thing another exception such as `ValueError` fits better here

rohmu/object_storage/s3.py
```diff
@@ -486,6 +488,26 @@ def get_file_size(self, key: str) -> int:
...
+            chunks = math.ceil(size / chunk_size)
```
giacomo-alzetta-aiven, line 507: Can't we just do:

```suggestion
        if chunks > S3_MAX_NUM_PARTS_PER_UPLOAD:
            chunk_size = math.ceil(size / S3_MAX_NUM_PARTS_PER_UPLOAD)
            if chunk_size > S3_MAX_PART_SIZE_BYTES:
                raise StorageError(f"Cannot upload a file of size {size}. Chunk size {chunk_size} is too big for each part of multipart upload.")
            chunks = math.ceil(size / chunk_size)
```

?

This would find the smallest chunk size that fits exactly in the maximum number of chunks without having to loop. The cons is that the chunk size won't be a "round number".

ali-gholami-aiven: good idea, applied it. I don't think there's an obvious advantage between doing slightly bigger chunk size vs more chunks with smaller size, so for simplicity I opted for your solution. we may re-evaluate this later on.

giacomo-alzetta-aiven (approved):
LGTM

PR #205, opened 2025-05-02
s3: ability to gradually increase chunk size for multipart uploard

ali-gholami-aiven:

Resolves: #xxxxx

# Why this way


rohmu/object_storage/s3.py
```diff
@@ -543,6 +555,7 @@ def multipart_upload_file_object(
         mp_id = cmu_response["UploadId"]
 
         while True:
+            chunk_size = self.get_multipart_chunk_size(current_chunk_size=chunk_size, current_part_number=part_number)
```
giacomo-alzetta-aiven, line 558: I think logging the chunk size change would make sense:

```suggestion
            old_chunk_size = chunk_size
            chunk_size = self.get_multipart_chunk_size(current_chunk_size=chunk_size, current_part_number=part_number)
            if chunk_size != old_chunk_size:
                self.log.info("Updated chunk size from %s to %s at part_number %s", old_chunk_size, chunk_size, part_number)
```

test/object_storage/test_s3.py
```diff
@@ -395,7 +400,7 @@ def test_calculate_chunks_and_chunk_size(
 ) -> None:
```
giacomo-alzetta-aiven, line 400: Should rename the test here

giacomo-alzetta-aiven (commented):
Overall looks good.

I think we should probably add these to other storage backends as well though.

giacomo-alzetta-aiven:
Closing this since we added #206

PR #210, opened 2025-10-10
s3: support OVH EU S3-compatible storage

giacomo-alzetta-aiven:

Handle `BucketAlreadyOwnedByYou` when trying to create a bucket in `S3Transfer` by ignoring it. If we find that the bucket already exists and is owned by us we should be good, we shouldn't bail out on transfer creation.

This specific situation currently seem to only happen in OVH EU S3-compatible Object Storage service implementation when a user only has `s3:ListBucket` permission with a condition attached.

Resolves: #211 

# Why this way

This is the simplest fix that allows use to use the OVH EU S3-compatible Object Storage service.

# Detailed description

The AWS S3 documentation for `HeadBucket` specifies:

> If the bucket doesn't exist or you don't have permission to access it,
> the `HEAD` request returns a generic `400 Bad Request`, `403 Forbidden`,
> or `404 Not Found` HTTP status code.
> A message body isn't included, so you can't determine the exception
> beyond these HTTP response codes.

Which is unfortunate since it is often the case that buckets has `s3:ListBucket` permission set with a condition to only allow listing under certain prefixes, which causes `HeadBucket` to fail.

This hasn't been a problem before because generally most S3-compatible storages will return `403 Forbidden` for a bucket that exists but for which we do not have `s3:ListBucket` permission or that permission is only allowed under a prefix.

We recently discovered that OVH EU S3-compatible Object Storage service instead returns a `400 Bad Request` which causes rohmu to assume the bucket doesn't exist and than the `CreateBucket` will fail with `BucketAlreadyOwnedByYou`.

So make sure that if we see that specific error code we do nto bail out and instead continue on since everything is actually good. We do not consider `BucketAlreadyExists` because that could mean the bucket exists on a different account which may be an error in the configuration.

Note that the behaviour for `create_if_needed=False` is unchanged and hence wrong for OVH EU S3-compatible storage. We will address it in the future likely falling back to `GetBucketLocation` and/or `GetBucketAcl` to determine if the bucket exists but we don't have the correct permissions for `HeadBucket`.

### DIFF
--- a/rohmu/object_storage/s3.py
+++ b/rohmu/object_storage/s3.py
@@ -715,7 +715,15 @@ class S3Transfer(BaseTransfer[Config]):
             }
 
         self.stats.operation(StorageOperation.create_bucket)
-        self.get_client().create_bucket(**args)
+        try:
+            self.get_client().create_bucket(**args)
+        except botocore.exceptions.ClientError as ex:
+            # Ignore BucketAlreadyOwnedByYou error since that means that:
+            # 1. The bucket exists and doesn't need to be created
+            # 2. We own it so we assume permissions have been setup properly
+            # bail out if we get any other error, including BucketAlreadyExists
+            if ex.response.get("Error", {}).get("Code") != "BucketAlreadyOwnedByYou":
+                raise
 
     def create_concurrent_upload(
         self,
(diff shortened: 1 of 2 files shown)


ali-gholami-aiven (approved):

ali-gholami-aiven (approved):

jadkik (approved):
LGTM

ali-gholami-aiven (approved):

PR #219, opened 2026-01-14
s3: fix multipart upload debug log

gqmelo:

The log would raise an exception whenever the total number of parts is unknown because it tried to format `<?>` as a number (introduced by https://github.com/Aiven-Open/rohmu/pull/215).

# Why this way

The issue could have been detected by the current tests if we weren't mocking the logger.
Pytest already offers better ways to capture the logs and assert expected logs, so use the `caplog` fixture instead.

Additionally I also switched the tests to use `tmp_path` fixture which is better than creating a regular temp file as it will be in dedicated dirs associated to test names.


giacomo-alzetta-aiven (approved):
LGTM!

PR #222, opened 2026-05-19
s3: add user_agent_extra configuration parameter

aiven-marko-rauhamaa:
Allow callers to inject a custom User-Agent suffix into the botocore session used by S3Transfer.

The motivation is AWS PRM attribution (ticket IP-876), but the Rohmu feature is generic and backward-compatible.

### DIFF
--- a/rohmu/object_storage/s3.py
+++ b/rohmu/object_storage/s3.py
@@ -137,6 +137,7 @@ class S3Transfer(BaseTransfer[Config]):
         statsd_info: Optional[StatsdConfig] = None,
         ensure_object_store_available: bool = True,
         min_multipart_chunk_size: Optional[int] = None,
+        user_agent_extra: Optional[str] = None,
     ) -> None:
         super().__init__(
             prefix=prefix,
@@ -161,6 +162,7 @@ class S3Transfer(BaseTransfer[Config]):
         self.use_dualstack_endpoint = use_dualstack_endpoint
         self.default_multipart_chunk_size = max(segment_size, min_multipart_chunk_size or 0)
         self.encrypted = encrypted
+        self.user_agent_extra = user_agent_extra
         self.s3_client: Optional[S3Client] = None
         self.location = ""
         if not self.host or not self.port:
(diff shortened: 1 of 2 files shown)


rohmu/object_storage/s3.py
```diff
@@ -217,6 +219,7 @@ def get_client(self) -> S3Client:
                 if self.use_dualstack_endpoint is True:
                     custom_config["use_dualstack_endpoint"] = True
                 with self._get_session() as session:
+                    session.user_agent_extra = self.user_agent_extra or ""
```
heikju: I think in a `S3Transfer` object we only construct a single `botocore.Session` instance, and only a single client, so this could be done in a single place in `_get_client()` inside the lock I think, because this is essentially part of session object construction, it's just not a constructor arg for some reason.

heikju (dismissed):
I would prefer constructing the session completely in a single place, especially since it seems to be protected by a lock.

rohmu/object_storage/s3.py
aiven-marko-rauhamaa: The singleton session object is independent of `self` (the transfer object). It is not obvious that the same `user_agent_extra` setting can be applied unconditionally to all API requests. That's why it is applied locally here.

heikju: Is there a guarantee then that modifying the globally shared session only has effect on the particular client you intend it to have effect on, instead of all or some of them if multiple S3Transfer instances are created and used in parallel? I had a look at botocore.Session implementation but it's really not obvious to me here.

aiven-marko-rauhamaa: It would have probably been ok, but now I look more closely, the proper place for the extra info is inside `config`. I've updated the PR accordingly and taken this change back to draft until I've tested it.

heikju: If there happens to be a good way to clearly set this per connection only, then that would be preferable. And if it turns out there isn't, then we can always think about if our use cases would be fulfilled with a single global session level setting that affects all clients.

aiven-marko-rauhamaa: The manual tests indicate it's working. I have made it reviewable again.

heikju (approved):
If configuring user-agent per client like this works without touching the global shared session, then good enough for me.

giacomo-alzetta-aiven (approved):
LGTM

PR #224, opened 2026-08-25
object_storage: add support for regional storage endpoints [PGSQL-1190]

tkren:

Instead of using the global storage endpoint https://storage.googleapis.com/storage/v1/ passing `direct_location=REGION_OR_MULTI_REGION` allows to use the regional endpoint instead.

See https://docs.cloud.google.com/storage/docs/request-endpoints and https://docs.cloud.google.com/storage/docs/regional-endpoints.

### DIFF
--- a/rohmu/object_storage/config.py
+++ b/rohmu/object_storage/config.py
@@ -12,6 +12,7 @@ from rohmu.common.models import ProxyInfo, StorageDriver, StorageModel
 from typing import Any, Dict, Final, Literal, Optional, TypeVar
 
 import platform
+import re
 
 StorageModelT = TypeVar("StorageModelT", bound=StorageModel)
 
@@ -58,6 +59,8 @@ AZURE_MAX_NUM_PARTS_PER_UPLOAD: Final[int] = 10000
 GOOGLE_DOWNLOAD_CHUNK_SIZE: Final[int] = 1024 * 1024 * 5 if (get_total_memory() or 0) < 2048 else 1024 * 1024 * 50
 GOOGLE_UPLOAD_CHUNK_SIZE: Final[int] = 1024 * 1024 * 5
 GOOGLE_MAX_NUM_PARTS_PER_UPLOAD: Final[int] = 10000
+# GCS region/location names as used in regional endpoint hostnames, e.g. "us-west4", "europe-west1", "nam4".
+GOOGLE_DIRECT_LOCATION_PATTERN: Final[str] = r"^[a-z0-9]+(-[a-z0-9]+)*$"
 
 LOCAL_CHUNK_SIZE: Final[int] = 1024 * 1024
 
@@ -123,6 +126,7 @@ class GoogleObjectStorageConfig(StorageModel):
     credential_file: Optional[Path] = None
     credentials: Optional[Dict[str, Any]] = Field(None, repr=False)
     proxy_info: Optional[ProxyInfo] = None
+    direct_location: Optional[str] = None
     prefix: Optional[str] = None
     storage_type: Literal[StorageDriver.google] = StorageDriver.google
 
@@ -132,6 +136,18 @@ class GoogleObjectStorageConfig(StorageModel):
         if values["project_id"] is None and values["bucket_name"] is None:
             raise ValueError("at least one of project_id, bucket_name must be set")
         return values
+
+    @validator("direct_location")
+    @classmethod
+    def valid_direct_location(cls, v: Optional[str]) -> Optional[str]:
+        # None means "use the global endpoint". Anything else is interpolated into a hostname,
+        # so reject malformed values here; "" would otherwise yield
+        # https://storage..rep.googleapis.com/storage/v1/
+        if v is None:
+            return v
+        if not re.fullmatch(GOOGLE_DIRECT_LOCATION_PATTERN, v):
+            raise ValueError(f"invalid direct_location: {v!r}, must match {GOOGLE_DIRECT_LOCATION_PATTERN}")
+        return v
 
 
 class LocalObjectStorageConfig(StorageModel):
--- a/rohmu/object_storage/google.py
+++ b/rohmu/object_storage/google.py
@@ -224,6 +224,7 @@ class GoogleTransfer(BaseTransfer[Config]):
         notifier: Optional[Notifier] = None,
         statsd_info: Optional[StatsdConfig] = None,
         ensure_object_store_available: bool = True,
+        direct_location: Optional[str] = None,
     ) -> None:
         super().__init__(
             prefix=prefix,
@@ -233,6 +234,16 @@ class GoogleTransfer(BaseTransfer[Config]):
         )
         self.project_id = project_id
         self.proxy_info = proxy_info
+
+        # default is using the global storage endpoint https://storage.googleapis.com/storage/v1/, but we can override it
+        # with the regional endpoint https://storage.{direct_location}.rep.googleapis.com/storage/v1/ here.
+        # see https://docs.cloud.google.com/storage/docs/regional-endpoints
+        self.regional_endpoint: str | None
+        if direct_location is not None:
+            self.regional_endpoint = f"https://storage.{direct_location}.rep.googleapis.com/storage/v1/"
+        else:
+            self.regional_endpoint = None
+
         self.google_creds = get_credentials(credential_file=credential_file, credentials=credentials)
         self.gs: Optional[StorageResource] = self._init_google_client()
         self.gs_object_client: Optional[StorageResource.ObjectsResource] = None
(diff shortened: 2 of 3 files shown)


glenn-mcallister-aiven (commented):
Two blockers, four non-blocking.

`test_region_endpoint` omits the `get_credentials` patch every sibling test applies, so `__init__` reaches `google.auth.default()` — `DefaultCredentialsError` on any runner without ADC.

`direct_location` reaches `__init__` only, not `GoogleObjectStorageConfig`, so no config-driven caller can set it. It fails silently rather than loudly: `StorageModel.Config` sets `extra_forbid = True`, not a pydantic v1 key (`extra`, cf. `rohmu/common/models.py:56`), so the extra is ignored and the transfer uses the global endpoint. The typo is pre-existing; the missing field is this PR's.

rohmu/object_storage/google.py
```diff
@@ -224,6 +224,7 @@ def __init__(
         notifier: Optional[Notifier] = None,
         statsd_info: Optional[StatsdConfig] = None,
         ensure_object_store_available: bool = True,
+        direct_location: Optional[str] = None,
```
glenn-mcallister-aiven, line 229: issue (blocking): `direct_location` absent from `GoogleObjectStorageConfig`

`get_transfer`/`get_transfer_model` construct via `config_model(**storage_config)`, so the parameter is unreachable from config. And it is dropped silently: `StorageModel.Config` sets `extra_forbid = True` — the pydantic v1 key is `extra` (`rohmu/common/models.py:56`), so extras are ignored, not forbidden. Add `direct_location: Optional[str] = None` to `GoogleObjectStorageConfig` (`config.py:119`). The `extra_forbid` typo is pre-existing — leave it.

rohmu/object_storage/google.py
```diff
@@ -233,6 +234,16 @@ def __init__(
         )
         self.project_id = project_id
         self.proxy_info = proxy_info
+
+        # default is using the global storage endpoint https://storage.googleapis.com/storage/v1/, but we can override it
+        # with the regional endpoint https://storage.{direct_location}.rep.googleapis.com/storage/v1/ here.
+        # see https://docs.cloud.google.com/storage/docs/regional-endpoints
+        self.regional_endpoint: str | None
+        if direct_location is not None:
+            self.regional_endpoint = f"https://storage.{direct_location}.rep.googleapis.com/storage/v1/"
```
glenn-mcallister-aiven, line 247: suggestion (non-blocking): validate `direct_location`

Unvalidated interpolation into a hostname. `""` yields `https://storage..rep.googleapis.com/storage/v1/`; a typo yields a resolvable-looking host. Static discovery defers the failure to the first request, where `_init_google_client` retries `ServerNotFoundError` for 600 s. A regex here fails immediately instead.

rohmu/object_storage/google.py
```diff
@@ -275,9 +286,12 @@ def _init_google_client(self) -> StorageResource:
             authorized_http = google_auth_httplib2.AuthorizedHttp(self.google_creds, http=http)
 
             try:
+                kwargs = {"http": authorized_http}
```
glenn-mcallister-aiven: suggestion (non-blocking): explicit `client_options` over `**kwargs`

    client_options = {"api_endpoint": self.regional_endpoint} if self.regional_endpoint else None
    return build("storage", "v1", http=authorized_http, client_options=client_options)

Shorter, and keeps the keywords visible to the checker. Confirm `make mypy` is green as written: `build()` in `google-api-python-client-stubs` is overloaded on `Literal["storage"], Literal["v1"]`, and `**kwargs` defeats overload resolution. Green CI settles it.

rohmu/object_storage/google.py
```diff
@@ -275,9 +286,12 @@ def _init_google_client(self) -> StorageResource:
...
+                if self.regional_endpoint is not None:
+                    kwargs["client_options"] = {"api_endpoint": self.regional_endpoint}
```
glenn-mcallister-aiven: question (non-blocking): batch requests and the regional endpoint

`client_options.api_endpoint` overrides `Resource._baseUrl`. `new_batch_http_request()` builds its URI from the discovery document's `rootUrl` + `batchPath`. If that holds, `delete_keys` (`google.py:539`) stays on `storage.googleapis.com` under `direct_location` — data residency lost, and probably a hard failure, since the `rep` endpoints likely do not serve the JSON API batch path. Unverified here: no `googleapiclient` in my checkout. One assertion on the batch URI settles it.

test/object_storage/test_google.py
```diff
@@ -692,3 +692,28 @@ def test_project_id_required_for_ensuring_object_store() -> None:
...
+    ("direct_location", "expected_endpoint"),
+    (
+        (None, "https://storage.googleapis.com/storage/v1/"),
+        ("us-west4", "https://storage.us-west4.rep.googleapis.com/storage/v1/"),
+    ),
+)
+def test_region_endpoint(direct_location: str | None, expected_endpoint: str | None) -> None:
+    """Test that we can create a transfer with a custom region endpoint."""
+    with ExitStack() as stack:
+        stack.enter_context(patch("rohmu.object_storage.google.GoogleTransfer._verify_object_storage_unwrapped"))
```
glenn-mcallister-aiven, line 714: issue (blocking): missing `get_credentials` patch

Every sibling test enters

    stack.enter_context(patch("rohmu.object_storage.google.get_credentials"))

This one does not, so `__init__` reaches `google.auth.default()` — `DefaultCredentialsError` without ADC.

test/object_storage/test_google.py
```diff
@@ -692,3 +692,28 @@ def test_project_id_required_for_ensuring_object_store() -> None:
...
+            bucket_name="test-bucket",
+            notifier=MagicMock(),
+            direct_location=direct_location,
+        )
+        if direct_location is None:
+            assert transfer.regional_endpoint is None
+        else:
+            assert transfer.regional_endpoint == expected_endpoint
+        assert transfer.gs is not None
+        assert transfer.gs._baseUrl == expected_endpoint  # type: ignore[attr-defined]
```
glenn-mcallister-aiven: thought (non-blocking): asserts a private googleapiclient attribute

`_baseUrl` is internal, so the assertion tracks library internals rather than behaviour. The URI of a built request is the observable equivalent.

rohmu/object_storage/google.py
```diff
@@ -233,6 +234,16 @@ def __init__(
         )
         self.project_id = project_id
         self.proxy_info = proxy_info
+
+        # default is using the global storage endpoint https://storage.googleapis.com/storage/v1/, but we can override it
```
glenn-mcallister-aiven, line 240: praise: link to the regional-endpoints docs

Both URL shapes plus the reference — readable without a lookup.

rohmu/object_storage/google.py
edward-evans-aiven: Documented on https://docs.cloud.google.com/storage/docs/regional-endpoints that the regional endpoint explicitly does not support JSON Batch operations. 

Batch requests are only used for myhoard deleting a stream, but that is an object per binlog needing to be deleted, which would now need to either round-trip for each object delete, or break when the global endpoint blocks us.

rohmu/object_storage/google.py
RommelLayco: added here https://github.com/Aiven-Open/rohmu/pull/224/changes#diff-748a9e28a5aa49926abcea03e50185958a9151089cc31f004fcbbb6888e3462eR129

rohmu/object_storage/config.py
```diff
@@ -133,6 +137,18 @@ def project_id_or_bucket_name_must_be_given(cls, values: Dict[str, Any]) -> Dict
...
 
+    @validator("direct_location")
+    @classmethod
+    def valid_direct_location(cls, v: Optional[str]) -> Optional[str]:
+        # None means "use the global endpoint". Anything else is interpolated into a hostname,
+        # so reject malformed values here; "" would otherwise yield
+        # https://storage..rep.googleapis.com/storage/v1/
+        if v is None:
+            return v
+        if not re.fullmatch(GOOGLE_DIRECT_LOCATION_PATTERN, v):
```
bugant, line 148: nit: It could be useful to compile the `GOOGLE_DIRECT_LOCATION_PATTERN` regexp.

bugant (approved):

egor-voynov-aiven (changes requested):
Since JSON batch operations don't work with regional endpoint we need add fallback for delete_keys method - delete keys one by one if regional_endpoint specified

RommelLayco (approved):
lgtm

egor-voynov-aiven:
> Since JSON batch operations don't work with regional endpoint we need add fallback for delete_keys method - delete keys one by one if regional_endpoint specified

We agreed to do it in separate PR

egor-voynov-aiven (approved):

PR #225, opened 2026-08-28
skip JSON batch API for GCS regional storage endpoints

egor-voynov-aiven:
Regional storage endpoints for Google Cloud Storage do not support the JSON API batch endpoint, so delete_keys() now falls back to deleting keys one by one when a regional endpoint is configured, instead of grouping delete requests into batch HTTP requests.


vladislav-manafov-aiven (approved):
LGTM

PR #226, opened 2026-08-30
compressor: bound DecompressSink output chunk size per algorithm [PGSQL-1193]

alexanderakimov-spec:
DecompressSink previously ran each algorithm's decompressor in a single unbounded call per write(), so a single compressed chunk could produce an arbitrarily large decompressed output in one shot. Split the decompression logic into per-algorithm sink classes (snappy/lzma/zstd) that cap output at OUTPUT_CHUNK_SIZE and slice snappy input at SNAPPY_INPUT_SLICE_SIZE, writing through _write_to_next_sink incrementally instead of decompressing the whole input in one call.

[PGSQL-1193]

### DIFF
--- a/rohmu/compressor.py
+++ b/rohmu/compressor.py
@@ -5,10 +5,11 @@
 from .errors import InvalidConfigurationError
 from .filewrap import Sink, Stream
 from .snappyfile import SnappyFile
-from .typing import BinaryData, Compressor, Decompressor, FileLike, HasRead, HasWrite
+from .typing import BinaryData, Compressor, FileLike, HasRead, HasWrite, SinkDecompressor
 from .zstdfile import open as zstd_open
 from typing import cast, IO
 
+import io
 import lzma
 
 try:
@@ -20,6 +21,10 @@ try:
     import zstandard as zstd
 except ImportError:
     zstd = None  # type: ignore
+
+
+OUTPUT_CHUNK_SIZE = 128 * 1024
+SNAPPY_INPUT_SLICE_SIZE = 256 * 1024
 
 
 def CompressionFile(dst_fp: FileLike, algorithm: str, level: int = 0, threads: int = 0) -> FileLike:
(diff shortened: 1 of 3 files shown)


egor-voynov-aiven (changes requested):
This abstraction is used in create_sink_pipeline. We need to make sure that the entire pipeline created by this function can operate in streaming mode.

Put together a simple script that decompresses and decrypts a highly compressible file (50–100 MB compressed, 1–2 GB uncompressed), and run it through create_sink_pipeline with the Memray memory profiler.

alexanderakimov-spec:
[memray-flamegraph-restore.html](https://github.com/user-attachments/files/31650009/memray-flamegraph-restore.html)
[memray-flamegraph-restore-FIXED.html](https://github.com/user-attachments/files/31650016/memray-flamegraph-restore-FIXED.html)
[restore_table_GOOGLE.html](https://github.com/user-attachments/files/31650018/restore_table_GOOGLE.html)
[restore_table_GOOGLE_FI
[Uploading fix_validation_report.md…]()
XED.html](https://github.com/user-attachments/files/31650019/restore_table_GOOGLE_FIXED.html)
Here is the collected reports from the memray and a summary, based on the reports

rohmu/compressor.py
```diff
@@ -78,25 +83,97 @@ def DecompressionFile(src_fp: FileLike, algorithm: str) -> FileLike:
     return src_fp
 
 
+class SnappySinkDecompressor(Sink):
```
egor-voynov-aiven: Why do the *SinkDecompressor classes inherit from Sink?

Sink declares a public write method, which, judging by the structure, is supposed to be used. However, it’s not used here and is not implemented correctly.

rohmu/compressor.py
```diff
@@ -78,25 +83,97 @@ def DecompressionFile(src_fp: FileLike, algorithm: str) -> FileLike:
...
+        io.RawIOBase.__init__(self)
+
+    def write(self, data: BinaryData) -> int:  # type: ignore[override]
+        return Sink.write(self, data)
+
+    def writable(self) -> bool:
+        return True
+
+
+class ZstdSinkDecompressor:
```
egor-voynov-aiven, line 136: Here you don't inherit from Sink, but naming same as for  LzmaSinkDecompressor and SnappySinkDecompressor. That is really confusing

rohmu/compressor.py
```diff
@@ -78,25 +83,97 @@ def DecompressionFile(src_fp: FileLike, algorithm: str) -> FileLike:
...
+            if output:
+                self._write_to_next_sink(output)
+
+            if self._decompressor.needs_input or self._decompressor.eof:
+                return
+
+            output = self._decompressor.decompress(
+                b"",
+                max_length=OUTPUT_CHUNK_SIZE,
+            )
```
egor-voynov-aiven, line 121: This part a bit confusing - why do we use while loop here?
We should definitely add explanation comment here

egor-voynov-aiven (changes requested):

rohmu/compressor.py
alexanderakimov-spec: We cannot inherit the Sink here -> we will introduce the cyclic referencing at stream_write(...). The helping class ZstdOutputSink was introduced because of this. The naming 'Sink' in the class name is because the decompressor is forwarding the output to the provided "Sink" directly, with no extra duplication

rohmu/compressor.py
alexanderakimov-spec: Yes, the need for the loop is that the input chunk (brought from the network) and the decompressor's output chunk are different by their sizes and we need to process the input chunk fully. I'll add the comment.

rohmu/compressor.py
alexanderakimov-spec: Because of the private method "_write_to_next_sink". Yes: we can call it, but this practice is not welcomed.

egor-voynov-aiven: But you haven't implemented Sink interface. If you need `_write_to_next_sink` it's not necessary to inherit from Sink

alexanderakimov-spec: Yes, it's only the matter of the public/private access: class Sink declares only the 'write' method as a public one. The '_write_to_next_sink' is not and callers are not expecting to use such methods directly (but could: Python makes the methods with `_` prefix  technically accessible)

egor-voynov-aiven: Just don't use Sink class for this. You inherited them from Sink, but it's actually not a Sink-like object

alexanderakimov-spec: Ok, I'll switch to direct usage of the `_write_to_next_sink`

egor-voynov-aiven (approved):
🚀

